PR #1, opened 2025-04-11
feat: add base for dictionary

tony-go:


binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  bare_ffmpeg_dictionary_t *dict;
+  err = js_create_arraybuffer(env, sizeof(bare_ffmpeg_dictionary_t), (void **) dict, &handle);
+
+  dict->handle = NULL;
+
+  return handle;
+}
+
+static js_value_t *
+bare_ffmpeg_dictionary_set(js_env_t *env, js_callback_info_t *info) {
```
kasperisager, line 1391:
```suggestion
bare_ffmpeg_dictionary_set_entry(js_env_t *env, js_callback_info_t *info) {
```

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_arraybuffer_info(env, argv[0], (void **) dict, NULL);
+  assert(err == 0);
+
+  size_t len;
+  err = js_get_value_string_utf8(env, argv[1], NULL, 0, &len);
+  assert(err == 0);
+
+  len += +1 /* NULL */;
+
+  utf8_t *key = malloc(len);
```
kasperisager, line 1411: Missing `free()`.

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_value_string_utf8(env, argv[1], key, len, NULL);
+  assert(err == 0);
+
+  err = js_get_value_string_utf8(env, argv[2], NULL, 0, &len);
+  assert(err == 0);
+
+  len += +1 /* NULL */;
+
+  utf8_t *value = malloc(len);
```
kasperisager, line 1420: Missing `free()`.

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  assert(err == 0);
+
+  err = av_dict_set(&dict->handle, (const char *) key, (const char *) value, 0);
+  assert(err);
+
+  return NULL;
+}
+
+static js_value_t *
+bare_ffmpeg_dictionary_get(js_env_t *env, js_callback_info_t *info) {
```
kasperisager, line 1431:
```suggestion
bare_ffmpeg_dictionary_get_entry(js_env_t *env, js_callback_info_t *info) {
```

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_arraybuffer_info(env, argv[0], (void **) dict, NULL);
+  assert(err == 0);
+
+  size_t len;
+  err = js_get_value_string_utf8(env, argv[1], NULL, 0, &len);
+  assert(err == 0);
+
+  len += +1 /* NULL */;
+
+  utf8_t *key = malloc(len);
```
kasperisager, line 1451: Missing `free()`.

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_value_string_utf8(env, argv[1], key, len, NULL);
+  assert(err == 0);
+
+  AVDictionaryEntry *entry = av_dict_get(dict->handle, (const char *) key, NULL, 0);
+  assert(entry != NULL);
+
+  js_value_t *result;
+  err = js_create_string_utf8(env, (const utf8_t *) entry->value, sizeof(entry->value), &result);
```
kasperisager, line 1459:
```suggestion
  err = js_create_string_utf8(env, (const utf8_t *) entry->value, -1, &result);
```

`sizeof(entry->value)` will give you the size of `char *`, i.e. the size of a pointer, and not the size of the string value.

binding.c
```diff
@@ -1369,6 +1374,113 @@ bare_ffmpeg_scaler_scale(js_env_t *env, js_callback_info_t *info) {
...
+  assert(argc == 1);
+
+  bare_ffmpeg_dictionary_t *dict;
+  err = js_get_arraybuffer_info(env, argv[0], (void **) dict, NULL);
+  assert(err == 0);
+
+  av_dict_free(&dict->handle);
+
+  return NULL;
+}
```
kasperisager, line 1482: Move above `bare_ffmpeg_dictionary_init()` so `*_destroy()` always immediately follows `*_init()`.

binding.c
```diff
@@ -1438,6 +1550,11 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   V("initScaler", bare_ffmpeg_scaler_init)
   V("destroyScaler", bare_ffmpeg_scaler_destroy)
   V("scaleScaler", bare_ffmpeg_scaler_scale)
+
+  V("initDictionary", bare_ffmpeg_dictionary_init)
+  V("dictionarySet", bare_ffmpeg_dictionary_get)
+  V("dictionaryGet", bare_ffmpeg_dictionary_set)
+  V("destroyDictionary", bare_ffmpeg_dictionary_destroy)
```
kasperisager, line 1557:
```suggestion
  V("initDictionary", bare_ffmpeg_dictionary_init)
  V("destroyDictionary", bare_ffmpeg_dictionary_destroy)
  V("getDictionaryEntry", bare_ffmpeg_dictionary_get_entry)
  V("setDictionaryEntry", bare_ffmpeg_dictionary_set_entry)
```

index.js
```diff
@@ -15,6 +15,7 @@ const Packet = require('./lib/packet')
 const Scaler = require('./lib/scaler')
 const Stream = require('./lib/stream')
 const Rational = require('./lib/rational')
+const Dictonary = require('./lib/dictionary.js')
```
kasperisager, line 18:
```suggestion
const Dictonary = require('./lib/dictionary')
```

lib/dictionary.js
```diff
@@ -0,0 +1,23 @@
+const binding = require('../binding')
+const ReferenceCounted = require('./reference-counted')
+
+module.exports = class Dictionary extends ReferenceCounted {
```
kasperisager, line 4:
```suggestion
module.exports = class FFmpegDictionary extends ReferenceCounted {
```

Qualified names makes it much easier to spot where the class is coming from when looking at debug output and heap traces.

lib/dictionary.js
```diff
@@ -0,0 +1,23 @@
...
+    this._handle = binding.initDictionary()
+  }
+
+  _destroy() {
+    binding.destroyDictionary(this._handle)
+    this._handle = null
+  }
+
+  set(key, value) {
+    return binding.dictionarySet(this._handle, key, value)
```
kasperisager, line 17:
```suggestion
    return binding.setDictionaryEntry(this._handle, key, value)
```

lib/dictionary.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  get(key) {
+    return binding.dictionaryGet(this._handle, key)
```
kasperisager, line 21:
```suggestion
    return binding.getDictionaryEntry(this._handle, key)
```

kasperisager (changes requested):

binding.c
tony-go: Yeah I confused with `strlen()` I guessed haha. Shot!

test.js
```diff
@@ -1,6 +1,14 @@
 const test = require('brittle')
 const ffmpeg = require('.')
 
+test('dictionary', (t) => {
```
tony-go, line 4: let me add more tests

package.json
```diff
@@ -16,7 +16,7 @@
   ],
   "addon": true,
   "scripts": {
-    "test": "prettier . --check && bare test.js"
+    "test": "prettier . --check && bare test/index.js"
```
kasperisager, line 19:
```suggestion
    "test": "prettier . --check && bare test.js"
```

Move `test/index.js` to `test.js` so we maintain the idiom that `test.js` is the single entry point for tests.

kasperisager (changes requested):

test/dictionary.js
```diff
@@ -0,0 +1,62 @@
...
+
+    t.exception.all(() => {
+      dict.set('foo', '')
+    }, 'Value should be a non empty string')
+  })
+
+  t.test('get throw when key is empty', (t) => {
+    const dict = new ffmpeg.Dictionary()
+
+    t.exception.all(() => {
```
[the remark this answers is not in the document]
kasperisager: "Native" means "builtin" in this case, e.g. `TypeError`, `RangeError`, `SyntaxError` et. al. Those usually indicate more serious issues, not application exceptions.

test/dictionary.js
```diff
@@ -0,0 +1,62 @@
...
+      dict.get('')
+    }, 'Key should be a non empty string')
+  })
```
kasperisager, line 61: Move these to the top-level instead, there's no need to use subtests here.

kasperisager (changes requested):

test/dictionary.js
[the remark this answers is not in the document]
tony-go: Oh I was not sure :) Thanks

test/dictionary.js
tony-go: I am really about this overlay:

```
# dictionary
    ok 1 - (basic set/get) - retrieves value after set
    ok 2 - (multiple entries) - should be equal
    ok 3 - (multiple entries) - should be equal
    ok 4 - (multiple entries) - should be equal
    ok 5 - (overwrite existing key) - last write wins
    ok 6 - (non-existent key returns null) - getting unknown key yields null
    ok 7 - (set throw when key is empty) - Key should be a non empty string
    ok 8 - (set throw when value is empty) - Value should be a non empty string
    ok 9 - (get throw when key is empty) - Key should be a non empty string
ok 5 - dictionary # time = 1ms
```

But as you prefer tbh

package.json
tony-go: make sense 👍🏼

test/dictionary.js
kasperisager: Subtests absolutely have their use, but they should only be reached for when necessary.

kasperisager (approved):

PR #2, opened 2025-04-15
feat: add input-format

tony-go:


binding.c
```diff
@@ -176,6 +181,50 @@ bare_ffmpeg_output_format_init(js_env_t *env, js_callback_info_t *info) {
   return handle;
 }
 
+static js_value_t *
+bare_ffmpeg_input_format_init(js_env_t *env, js_callback_info_t *info) {
+  avdevice_register_all();
```
tony-go, line 186: I would like to create it's own js function for this use case, but not until we have another case than this one. Am I wrong? What is your tought on that @kasperisager

cmake/ports/ffmpeg/port.cmake
```diff
@@ -71,7 +71,6 @@ if(APPLE)
     "--sysroot=${CMAKE_OSX_SYSROOT}"
 
     --enable-avfoundation
-    --enable-coreimage
```
tony-go, line 74: With I got a 

```
Uncaught Error: dlopen(/Users/tonygo/bare/bare-ffmpeg/prebuilds/darwin-arm64/bare-ffmpeg.bare, 0x0005): symbol not found in flat namespace '_OBJC_CLASS_$_NSOpenGLPixelFormat'
```

Still have to figured out why ...

tony-go: Let me go through to see if I can grab any pointers: https://trac.ffmpeg.org/wiki/CompilationGuide/macOS

tony-go: This flag is not used here for instance: https://github.com/homebrew-ffmpeg/homebrew-ffmpeg/blob/master/Formula/ffmpeg.rb

binding.c
tony-go: But at the same time function will be call for each `bare_ffmpeg_input_format_init` which does not make sense as it should be called once before call of `bare_ffmpeg_input_format_init`

kasperisager: This should be run in a `uv_once_t` callback triggered from `bare_ffmpeg_exports()`. You can see an example here: https://github.com/holepunchto/bare-os/blob/b2dcff223cd6aa0f9dd99e6379c74df365aed9a9/binding.c#L10-L16

cmake/ports/ffmpeg/port.cmake
kasperisager: See https://github.com/FFmpeg/FFmpeg/blob/a888975a3c25760027cd59932f5c1ad04368db8b/configure#L203. We aren't enabling AppKit so that's likely why it's failing now that it's starting to use an AppKit symbol (https://developer.apple.com/documentation/appkit/nsopenglpixelformat). If everything runs fine without CoreImage then all good 👍

tony-go: > We aren't enabling AppKi

It sounds like we enabled appkit right? https://github.com/holepunchto/bare-ffmpeg/blob/main/cmake/ports/ffmpeg/port.cmake#L80

test/input-format.js
```diff
@@ -0,0 +1,14 @@
+const test = require('brittle')
+const ffmpeg = require('..')
+
+test('input format should wrap an array buffer', (t) => {
+  const inputFormat = new ffmpeg.InputFormat('avfoundation')
```
tony-go, line 5: I should handle each input format according to the os.

tony-go: Also it will be hard to run on CI, I guess

```text
Uncaught Error: dlopen(/Users/runner/work/bare-ffmpeg/bare-ffmpeg/prebuilds/darwin-arm64/bare-ffmpeg.bare, 0x0005): symbol not found in flat namespace '_AVCaptureDeviceTypeBuiltInMicrophone'
```

cmake/ports/ffmpeg/port.cmake
kasperisager: Ah, we are for macOS, yes!

test/input-format.js
```diff
@@ -0,0 +1,36 @@
...
+test('input format should wrap an array buffer', (t) => {
+  let inputFormat
+
+  if (Bare.platform == 'darwin') {
+    inputFormat = new ffmpeg.InputFormat('avfoundation')
+  } else if (Bare.platform == 'linux') {
+    inputFormat = new ffmpeg.InputFormat('v4l2')
+  } else if (Bare.platform == 'win32') {
+    inputFormat = new ffmpeg.InputFormat('dshow')
+  } else {
```
tony-go, line 13: It does not make sense as the next assertion will run anyway 🫣

test/input-format.js
```diff
@@ -0,0 +1,36 @@
...
+
+test('input format should wrap an array buffer', (t) => {
+  let inputFormat
+
+  if (Bare.platform == 'darwin') {
+    inputFormat = new ffmpeg.InputFormat('avfoundation')
+  } else if (Bare.platform == 'linux') {
+    inputFormat = new ffmpeg.InputFormat('v4l2')
+  } else if (Bare.platform == 'win32') {
+    inputFormat = new ffmpeg.InputFormat('dshow')
```
kasperisager, line 12: Let's make this the default if no name is given.

kasperisager (changes requested):

test/input-format.js
```diff
@@ -0,0 +1,36 @@
...
+  } else {
+    t.pass()
+  }
+
+  t.ok(inputFormat._handle instanceof ArrayBuffer)
```
kasperisager, line 17:
```suggestion
```

Private API.

kasperisager (changes requested):

test/input-format.js
tony-go: I proposed something different.

test/input-format.js
kasperisager: @tony-go Ah, I meant default as in have `new ffmpeg.InputFormat()` pick the right one for the current platform.

kasperisager (changes requested):

test/input-format.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (Bare.platform == 'darwin') {
+    inputFormat = new ffmpeg.InputFormat('avfoundation')
+  } else if (Bare.platform == 'linux') {
+    inputFormat = new ffmpeg.InputFormat('v4l2')
+  } else if (Bare.platform == 'win32') {
+    inputFormat = new ffmpeg.InputFormat('dshow')
+  }
+
+  t.ok(inputFormat instanceof ffmpeg.InputFormat)
```
kasperisager, line 15:
```suggestion
  t.ok(inputFormat)
```

Should be sufficient.

test/input-format.js
tony-go: Sorry I don't gotcha.

kasperisager: No worries, my first comment wasn't terribly clear!

kasperisager: My bad, I missed https://github.com/holepunchto/bare-ffmpeg/pull/2#discussion_r2046783961. I'll tweak it, just a minute.

kasperisager: Like this: c537b88

kasperisager (approved):

PR #4, opened 2025-04-17
feat: handle video with avformatinput

tony-go:


tony-go:
Note for later: remove all these warnings

https://github.com/holepunchto/bare-ffmpeg/actions/runs/14574152736/job/40876667502?pr=4#step:11:34

test/codec-context.js
```diff
@@ -0,0 +1,118 @@
+const test = require('brittle')
+const ffmpeg = require('..')
+
+test('codec context could be open without options', (t) => {
+  const codec = new ffmpeg.Codec('h264')
```
tony-go, line 5: Let see if we cannot use a default cross platform codec ...

lib/constants.js
```diff
@@ -6,7 +6,12 @@ module.exports = exports = {
   },
   pixelFormats: {
     RGBA: binding.AV_PIX_FMT_RGBA,
-    YUVJ420P: binding.AV_PIX_FMT_YUVJ420P
+    YUVJ420P: binding.AV_PIX_FMT_YUVJ420P,
+    UYVY422: binding.AV_PIX_FMT_UYVY422,
+    YUV420P: binding.AV_PIX_FMT_YUV420P
+  },
+  encoderMap: {
+    h264: 'libx264'
```
tony-go, line 14: The idea is that if the user choose `h264` we would pick the right encoder for him.

lib/playback.js
```diff
@@ -0,0 +1,28 @@
+const binding = require('../binding')
+const ReferenceCounted = require('./reference-counted')
+
+module.exports = class Playback extends ReferenceCounted {
```
tony-go, line 4: This is temporary :) Until @mafintosh did its e2e test :)

tony-go: The whole cmake stuff added are broken too hahaha :)

tony-go:
We can def close that PR, as we have this one now https://github.com/holepunchto/bare-sdl/pull/7

PR #6, opened 2025-04-29
feat(codec): be able to find decoder by name

tony-go:


test.js
```diff
@@ -1,3 +1,4 @@
+require('./test/codec')
```
tony-go, line 1: I use code abstract to test encoder and decoder as the encoder and decoder class are really thin :)

kasperisager (changes requested):
I don't think we need this. All codecs expose a numeric ID which is what we currently use for exposing static `Codec` instances for each supported codec, such as:

https://github.com/holepunchto/bare-ffmpeg/blob/1460480303b41b1191cfc7ebb0c7c3ac65fd6144/lib/codec.js#L37

For MPEG4, for example, we'd do:

```js
static MPEG4 = this.for(constants.codecs.MPEG4)
```

Have you encountered a case not supported by this, i.e. where a lookup by name is strictly required?

tony-go:
> Have you encountered a case not supported by this, i.e. where a lookup by name is strictly required?

@kasperisager Interesting. 

The code you shared imply that the encoder and decoder are the same right?

When I experimented with `av1` codec, I tried a bunch of encoder and decoder with different names.

- https://github.com/tony-go/av1/blob/code-alt/main.c#L101
- https://github.com/tony-go/av1/blob/code-alt/main.c#L224

So I am not sure when we put something generic, like `mpeg4` which implementation are picked.

My end goal was this one:

- https://github.com/holepunchto/bare-ffmpeg/pull/4/files#diff-56305c858ca1ee93a014f69fee9f6e71f1790bdf8c725a38dd17cb846df843ceR10
- https://github.com/holepunchto/bare-ffmpeg/pull/4/files#diff-602f56bb692d9bf3bdc5fef3d75d5fa33fe841c3d21ca46294bb6fdaf0785d0cR15

The user use a generic codec name and we picked the implementation for him.

So, I am totally OKAY to have a:

```js
 static H264 = this.for(constants.codecs.H264) 
``` 

if internally we have the power to pick up which implementation we use for the encoder an decoder.

kasperisager:
That was my previous point about us being selective about the implementations we support. The `AV_CODEC_ID_AV1` codec ID, for example, should refer to whichever encoder and decoder implementation we enable, such as `rav1e` and `dav1d`.

tony-go:
> That was my previous point about us being selective about the implementations we support. The `AV_CODEC_ID_AV1` codec ID, for example, should refer to whichever encoder and decoder implementation we enable, such as `rav1e` and `dav1d`.

Oh yeah with `--enable-*` whatever it is?

tony-go:
Let me close that one and try to do it on #4, it should save us a bunch of bits then :)

PR #9, opened 2025-05-15
feat: enhance InputFormatContext to take a InputFormat

tony-go:


lib/format-context.js
```diff
@@ -2,18 +2,24 @@ const binding = require('../binding')
...
+const InputFormat = require('./input-format')
 
 class FFmpegFormatContext extends ReferenceCounted {
   constructor(io) {
     super()
 
-    this._io = io._ref()
+    if (io) {
+      this._io = io._ref()
+    }
```
kasperisager, line 14:
```suggestion
    this._io = io ? io._ref() : null
```

This way the object shape doesn't change.

test/format-context.js
```diff
@@ -0,0 +1,54 @@
...
+  fallbackURL = 'testsrc=size=640x480:rate=30'
+}
+
+test('InputFormatContext should be instanciate with IOContext', (t) => {
+  const image = require('./fixtures/image/sample.jpeg', {
+    with: { type: 'binary' }
+  })
+  const io = new ffmpeg.IOContext(image)
+
+  const inputFormatContext = new ffmpeg.InputFormatContext(io)
```
kasperisager, line 18: This and the ones below should be destroyed on test teardown.

binding.c
```diff
@@ -401,6 +464,37 @@ bare_ffmpeg_format_context_get_streams(js_env_t *env, js_callback_info_t *info)
...
+  err = js_get_callback_info(env, info, &argc, argv, NULL, NULL);
+  assert(err == 0);
+
+  assert(argc == 1);
+
+  bare_ffmpeg_format_context_t *context;
+  err = js_get_arraybuffer_info(env, argv[0], (void **) &context, NULL);
+  assert(err == 0);
+
+  int stream_index = av_find_best_stream(context->handle, AVMEDIA_TYPE_VIDEO, -1, -1, NULL, 0);
```
kasperisager, line 483: Let's make this accept at least the type, e.g. `AVMEDIA_TYPE_VIDEO`, as an argument and then expose the types we need.

binding.c
```diff
@@ -401,6 +464,37 @@ bare_ffmpeg_format_context_get_streams(js_env_t *env, js_callback_info_t *info)
...
+  if (stream_index < 0) {
+    err = js_throw_error(env, NULL, "Best stream not found");
+    assert(err == 0);
+
+    return NULL;
+  }
+
+  js_value_t *result;
+  err = js_create_int32(env, stream_index, &result);
+  assert(err == 0);
```
kasperisager, line 493: This should return `-1` when no appropriate stream is found and then the caller can decide whether that's an error or not.

kasperisager (changes requested):

binding.c
```diff
@@ -401,6 +464,38 @@ bare_ffmpeg_format_context_get_streams(js_env_t *env, js_callback_info_t *info)
...
+  err = js_get_arraybuffer_info(env, argv[0], (void **) &context, NULL);
+  assert(err == 0);
+
+  int type;
+  err = js_get_value_int32(env, argv[1], &type);
+  assert(err == 0);
+
+  int stream_index = av_find_best_stream(context->handle, type, -1, -1, NULL, 0);
+  if (stream_index < 0) {
+    stream_index = -1;
```
tony-go, line 489: The doc describe that we could have two error cases:

```
 * @return  the non-negative stream number in case of success,
 *          AVERROR_STREAM_NOT_FOUND if no stream with the requested type
 *          could be found,
 *          AVERROR_DECODER_NOT_FOUND if streams were found but no decoder
```

Not sure we want that level of precision.

test/format-context.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const bestStream = inputFormatContext.getBestStream(
+    ffmpeg.constants.mediaType.SUBTITLE
+  )
+
+  t.is(bestStream, null)
+})
+
+function getOptions() {
+  const options = new ffmpeg.Dictionary()
```
tony-go, line 87: I also discovered that as far as we passed the dictionary to `InputFormatContext`'s constructor. The dictionary is destroyed 😨

binding.c
```diff
@@ -280,6 +280,69 @@ bare_ffmpeg_format_context_open_input(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_arraybuffer_info(env, argv[1], (void **) &options, NULL);
+  assert(err == 0);
+
+  size_t len;
+  err = js_get_value_string_utf8(env, argv[2], NULL, 0, &len);
+  assert(err == 0);
+
+  len += +1 /* NULL */;
+
+  utf8_t *url = malloc(len);
```
kasperisager, line 317: Missing a `free()`.

binding.c
kasperisager: The current approach seems fine and then we can always adjust as necessary 👍

lib/format-context.js
```diff
@@ -36,15 +40,60 @@ class FFmpegFormatContext extends ReferenceCounted {
...
+
+    if (bestIndex < 0) {
+      return null
+    }
+
+    if (this._streams[bestIndex]) {
+      return this._streams[bestIndex]
+    }
+
+    return null
```
kasperisager, line 63:
```suggestion
    if (bestIndex < 0 || bestIndex >= this._streams.length) {
      return null
    }

    return this._streams[bestIndex]
```

test/format-context.js
kasperisager: That's the documented behavior, yep. It might not make sense for `Dictionary` to be a reference counted type if it's always consumed when passed to another API function.

kasperisager (changes requested):

binding.c
tony-go: 😨

binding.c
```diff
@@ -319,6 +319,8 @@ bare_ffmpeg_format_context_open_input_with_format(js_env_t *env, js_callback_inf
   assert(err == 0);
 
   err = avformat_open_input(&context->handle, (char *) url, format->handle, &options->handle);
+  assert(err == 0);
```
kasperisager, line 322:
```suggestion

```

Should be allowed to fail.

kasperisager (changes requested):

binding.c
tony-go: I added it back mechanically, but I guess that I did not added it on purpose ... haha

tony-go: Sorry for my lack of focus haha!

kasperisager: All good! 😄

kasperisager (approved):

PR #10, opened 2025-05-19
feat(packet): add init from buffer some getters

tony-go:


lib/packet.js
```diff
@@ -2,9 +2,14 @@ const binding = require('../binding')
...
-  constructor() {
+  constructor(buffer) {
     super()
 
+    if (buffer instanceof ArrayBuffer) {
+      this._handle = binding.initPacketFromBuffer(buffer)
+      return
+    }
+
     this._handle = binding.initPacket()
```
kasperisager, line 13:
```suggestion
    if (buffer) {
      this._handle = binding.initPacketFromBuffer(buffer)
    } else {
      this._handle = binding.initPacket()
    }
```

binding.c
```diff
@@ -1558,11 +1593,60 @@ bare_ffmpeg_packet_unref(js_env_t *env, js_callback_info_t *info) {
...
+
+  js_value_t *result;
+  err = js_create_int32(env, packet->handle->stream_index, &result);
+  assert(err == 0);
+
+  return result;
+}
+
+static js_value_t *
+bare_ffmpeg_packet_get_data_as_array_buffer(js_env_t *env, js_callback_info_t *info) {
```
kasperisager, line 1625:
```suggestion
bare_ffmpeg_packet_get_data(js_env_t *env, js_callback_info_t *info) {
```

binding.c
```diff
@@ -1882,8 +1966,11 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   V("fillImage", bare_ffmpeg_image_fill)
 
   V("initPacket", bare_ffmpeg_packet_init)
+  V("initPacketFromBuffer", bare_ffmpeg_packet_init_from_buffer)
   V("destroyPacket", bare_ffmpeg_packet_destroy)
   V("unrefPacket", bare_ffmpeg_packet_unref)
+  V("getPacketStreamIndex", bare_ffmpeg_packet_get_stream_index)
+  V("getPacketDataAsArrayBuffer", bare_ffmpeg_packet_get_data_as_array_buffer)
```
kasperisager, line 1973:
```suggestion
  V("getPacketData", bare_ffmpeg_packet_get_data)
```

lib/packet.js
```diff
@@ -17,4 +22,12 @@ module.exports = class FFmpegPacket extends ReferenceCounted {
     this._unref()
     binding.unrefPacket(this._handle)
   }
+
+  get streamIndex() {
+    return binding.getPacketStreamIndex(this._handle)
+  }
+
+  get buffer() {
+    return binding.getPacketDataAsArrayBuffer(this._handle)
```
kasperisager, line 31:
```suggestion
  get data() {
    return Buffer.from(binding.getPacketData(this._handle))
```

kasperisager (changes requested):

lib/packet.js
tony-go: I really like this minimalism :)

lib/packet.js
kasperisager: We should also unpack the buffer so callers can provide a `Buffer` and not a raw `ArrayBuffer`.

kasperisager (changes requested):

lib/packet.js
```diff
@@ -2,10 +2,19 @@ const binding = require('../binding')
...
     super()
 
-    this._handle = binding.initPacket()
+    if (buffer) {
+      if (!buffer.buffer || !(buffer.buffer instanceof ArrayBuffer)) {
+        throw TypeError(
+          `Buffer parameter should be of type Buffer but got: ${typeof buffer}`
+        )
+      }
+      this._handle = binding.initPacketFromBuffer(buffer.buffer)
```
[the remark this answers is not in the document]
kasperisager: That's the correct approach! Remember to pass the byte offset and length as well.

lib/packet.js
```diff
@@ -5,12 +5,16 @@ module.exports = class FFmpegPacket extends ReferenceCounted {
   constructor(buffer) {
     super()
 
-    if (buffer instanceof ArrayBuffer) {
-      this._handle = binding.initPacketFromBuffer(buffer)
-      return
+    if (buffer) {
+      if (!buffer.buffer || !(buffer.buffer instanceof ArrayBuffer)) {
```
kasperisager, line 9:
```suggestion
      if (!Buffer.isBuffer(buffer)) {
```

kasperisager (changes requested):

lib/packet.js
```diff
@@ -17,4 +26,12 @@ module.exports = class FFmpegPacket extends ReferenceCounted {
     this._unref()
     binding.unrefPacket(this._handle)
   }
+
+  get streamIndex() {
+    return binding.getPacketStreamIndex(this._handle)
+  }
+
+  get buffer() {
```
kasperisager, line 34:
```suggestion
  get data() {
```

lib/packet.js
tony-go: Oh yeahhhhhhhhhhhhhhhhhhhhhh

kasperisager (changes requested):
https://github.com/holepunchto/bare-ffmpeg/pull/10#discussion_r2095712591

lib/packet.js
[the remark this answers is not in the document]
tony-go: Just to be sure I got it. We do it because we want to avoid this extra step?

```c
  size_t len;
  uint8_t *data;
  err = js_get_arraybuffer_info(env, argv[0], (void **) &data, &len);
```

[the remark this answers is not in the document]
tony-go: Also I am not sure what I've to do with the offset...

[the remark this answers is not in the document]
tony-go: Probably in case I want just a subpart on a Buffer?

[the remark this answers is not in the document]
kasperisager: Yep, an `ArrayBuffer` is a single, contiguous allocation and a `Buffer` as well as any other `TypedArray` is a view over either the entire or a portion of that allocation. When you access the `.buffer` property you're getting a pointer to the beginning of the allocation, not just the part that the `Buffer` or `TypedArray` corresponds to.

In summary, you need to pass along `.offset` and `.byteLength` to the native code as well and not use the length returned by `js_get_arraybuffer_info()` as that's the length of the entire allocation. `data` will point to the beginning of the allocation so to get the part that corresponds to the `Buffer` you should do:

```c
&data[offset]
```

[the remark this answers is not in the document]
kasperisager: See https://github.com/holepunchto/bare-buffer/blob/d7360379820e8a528b5bb2b7bf5a71221f3a36c1/index.js#L604-L613 for an example.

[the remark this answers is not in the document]
tony-go: Crystal clear!

kasperisager (approved):

PR #11, opened 2025-05-19
feat(codec-context): handle open with options

tony-go:


lib/codec-context.js
```diff
@@ -52,10 +52,14 @@ module.exports = class FFmpegCodecContext extends ReferenceCounted {
     binding.setCodecContextTimeBase(this._handle, view.buffer)
   }
 
-  open() {
+  open(options) {
     if (this._opened) return
     this._opened = true
-    binding.openCodecContext(this._handle)
+    if (options) {
+      binding.openCodecContext(this._handle, options._handle)
```
kasperisager, line 59: Let's add a `WithOptions` variant of this function to avoid argument juggling in native code. That'll make it much simpler to move to https://github.com/holepunchto/libjstl and the functions are then readily optimisable as well.

kasperisager (changes requested):

lib/codec-context.js
tony-go: Yeah, plus this is the pattern we followed so far

kasperisager (approved):

PR #12, opened 2025-05-20
feat(codec-context): add sendFrame 

tony-go:


binding.c
```diff
@@ -1024,6 +1024,48 @@ bare_ffmpeg_codec_context_send_packet(js_env_t *env, js_callback_info_t *info) {
...
+  err = js_get_arraybuffer_info(env, argv[0], (void **) &context, NULL);
+  assert(err == 0);
+
+  bare_ffmpeg_frame_t *frame;
+  err = js_get_arraybuffer_info(env, argv[1], (void **) &frame, NULL);
+  assert(err == 0);
+
+  err = avcodec_send_frame(context->handle, frame->handle);
+
+  if (err == AVERROR(EAGAIN) || err == AVERROR_EOF) {
```
tony-go, line 1048: I did not now how to handle `AVERROR_EOF` at the moment.

A potential approach could be to flush it by passing a `NULL` frame to `avcodec_send_frame` and then send a new frame. 

We should potentially expose a `flush` method to `CodeContext` but not sure if we really need it.

binding.c
```diff
@@ -1024,6 +1024,48 @@ bare_ffmpeg_codec_context_send_packet(js_env_t *env, js_callback_info_t *info) {
...
+    assert(err == 0);
+
+    return NULL;
+  }
+
+  js_value_t *result;
+  err = js_get_boolean(env, err == 0, &result);
+  assert(err == 0);
+
+  return result;
```
tony-go, line 1066: I also know that in the other `send*` method we expose we did not handle it that way, but I think that we should keep the hands on it. It could let the user to do retries etc..

binding.c
kasperisager: We'll eventually need to handle flushing as documented in https://ffmpeg.org/doxygen/7.0/group__lavc__encdec.html, yep.

binding.c
kasperisager: As mentioned above we should eventually handle flushing, which is currently missing.

kasperisager (approved):

binding.c
tony-go: I'll do it in a subsequent PR

PR #16, opened 2025-05-28
feat(frame): add .data getter

tony-go:


binding.c
```diff
@@ -1535,6 +1535,49 @@ bare_ffmpeg_frame_alloc(js_env_t *env, js_callback_info_t *info) {
...
+  bare_ffmpeg_frame_t *frame;
+  err = js_get_arraybuffer_info(env, argv[0], (void **) &frame, NULL);
+  assert(err == 0);
+
+  if (frame->handle->format != AV_PIX_FMT_RGBA &&
+      frame->handle->format != AV_PIX_FMT_RGB24) {
+    err = js_throw_error(env, NULL, "Unsupported pixel format");
+    assert(err == 0);
+    return NULL;
+  }
```
tony-go, line 1558: I think we should design a ds that could support multiple plane pixel format.

kasperisager (changes requested):
We should expose something backed by `av_image_copy_to_buffer()` instead as that should handle all the pixel format shenanigans.

PR #21, opened 2025-06-09
chore: scaflold libjstl

tony-go:


kasperisager (changes requested):
All casts should use an appropriate C++ cast operator such as `reinterpret_cast()`.

Issue #22, opened 2025-06-10
chore: port binding to `libjstl`


tony-go:

sethvincent:
I think we finished this with #23 unless there's anything else? All the function signatures look to be ported over.

PR #25, opened 2025-06-11
refactor(output-format, input-format, format-ctx): port to jstl

tony-go:
ref: #22

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -1,5 +1,4 @@
 #include <assert.h>
-#include <cstdint>
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>


binding.cc
```diff
@@ -115,270 +115,163 @@ bare_ffmpeg_io_context_destroy(js_env_t *env, js_receiver_t, js_arraybuffer_span
...
 
   if (format == NULL) {
-    err = js_throw_errorf(env, NULL, "No output format found for name '%s'", name);
+    err = js_throw_errorf(env, NULL, "No output format found for name '%s'", name.c_str());
     assert(err == 0);
 
-    free(name);
-
-    return NULL;
+    throw err;
```
kasperisager, line 128:
```suggestion
    throw js_pending_exception;
```

Otherwise it'll do `throw 0`, which currently works but only because `libjstl` isn't strict about it yet.

binding.cc
```diff
@@ -387,54 +280,32 @@ bare_ffmpeg_format_context_open_output(js_env_t *env, js_callback_info_t *info)
...
+  js_arraybuffer_span_of_t<bare_ffmpeg_format_context_t, 1> context
+) {
   int err;
 
-  size_t argc = 1;
-  js_value_t *argv[1];
-
-  err = js_get_callback_info(env, info, &argc, argv, NULL, NULL);
+  js_array_t result;
+  err = js_create_array(env, result);
```
kasperisager, line 301:
```suggestion
  uint32_t len = context->handle->nb_streams;

  js_array_t result;
  err = js_create_array(env, len, result);
```

binding.cc
```diff
@@ -387,54 +280,32 @@ bare_ffmpeg_format_context_open_output(js_env_t *env, js_callback_info_t *info)
...
-
-  uint32_t len = context->handle->nb_streams;
-
-  js_value_t *result;
-  err = js_create_array_with_length(env, len, &result);
-  assert(err == 0);
-
-  for (uint32_t i = 0; i < len; i++) {
-    js_value_t *handle;
+  for (uint32_t i = 0; i < context->handle->nb_streams; i++) {
```
kasperisager, line 304:
```suggestion
  for (uint32_t i = 0; i < len; i++) {
```

kasperisager (changes requested):

binding.cc
tony-go: Just out of curiosity, Why should we use length in that case? (We did not for array buffer)

kasperisager: It's only an optimization as we know the array length up front so we can do the equivalent of `new Array(len)`.

binding.cc
```diff
@@ -1,4 +1,5 @@
 #include <assert.h>
+#include <cstdint>
```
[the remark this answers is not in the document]
tony-go: yeah! I should be able to remove it

PR #35, opened 2025-06-17
chore(build): use cmake-harden

tony-go:


kasperisager (changes requested):
It needs this after `add_bare_module()`:

```cmake
harden(${bare_ffmpeg})
```

tony-go:
> It needs this after `add_bare_module()`:
> 
> ```cmake
> harden(${bare_ffmpeg})
> ```

whaaaat hahahahah

I should take a cup of tea and read https://github.com/holepunchto/bare-ffmpeg/pull/35

tony-go:
Now I have warning 🎉

tony-go:
Latest point to 14 too ... but it worked!

kasperisager:
I made the flags conditional on compiler support: https://github.com/holepunchto/cmake-harden/commit/9bafdd901c4761b2c39b3e6227c100bd04aeb856

kasperisager (changes requested):
I've added `cmake-harden` to `libjstl` in https://github.com/holepunchto/libjstl/commit/096669a07fbc70210f9e3b05739f15442b688801, let's update and see how many warnings are left after that.

tony-go:
> So far, so good! Let's fix the warnings as well so the slate is clean. I've yet to enable treating warnings as errors, but that will come eventually.

I was about to do into another PR, but yeah lessdoit

kasperisager (approved):

PR #37, opened 2025-06-17
libopus feature for ffmpeg

sethvincent:


CMakeLists.txt
```diff
@@ -6,7 +6,15 @@ find_package(cmake-ports REQUIRED PATHS node_modules/cmake-ports)
 
 project(bare_ffmpeg C CXX)
 
+bare_arch(arch)
+
+if(arch MATCHES "arm64")
```
sethvincent, line 11: This may need a more precise check. The cmake config in libopus doesn't seem to detect neon.

CMakeLists.txt
```diff
@@ -6,7 +6,15 @@ find_package(cmake-ports REQUIRED PATHS node_modules/cmake-ports)
...
+  set(OPUS_MAY_HAVE_NEON OFF CACHE BOOL "" FORCE)
+  set(OPUS_PRESUME_NEON ON CACHE BOOL "" FORCE)
+endif()
+
 fetch_package("github:holepunchto/libjstl#096669a")
+fetch_package("github:xiph/opus#c0eb2ca")
 
 find_port(ffmpeg FEATURES dav1d svt-av1 x264)
```
sethvincent, line 19:
```suggestion
find_port(ffmpeg FEATURES dav1d svt-av1 x264 libopus)
```

kasperisager (changes requested):
The `libopus` build definition must be in a separate port, see https://github.com/holepunchto/bare-ffmpeg/blob/main/cmake/ports/svt-av1/port.cmake for an example. It's fortunately built with CMake as well so the port should be trivial

kasperisager:
> Is there a way around recreating the equivalent of bare_arch for checking the arch?

That shouldn't be necessary at all. What's the issue you're running into?

tony-go (commented):
Looks good % @kasperisager comment! :)

sethvincent:
> > Is there a way around recreating the equivalent of bare_arch for checking the arch?
> 
> That shouldn't be necessary at all. What's the issue you're running into?

Ah, I see, I guess I was assuming because of this in the x264 port, which we could probably remove? https://github.com/holepunchto/bare-ffmpeg/blob/564d5b4038d166430252bd6e95f751589c60e20d/cmake/ports/x264/port.cmake#L23-L65

I got a little too into figuring out how the libopus build works with neon but I figured out the right option to make it build.

Assuming the workflow jobs succeed I think this is probably good now!

sethvincent:
> Assuming the workflow jobs succeed I think this is probably good now!

Narrator voice: "But not all the workflow jobs would succeed."

Now they do! I got really into getting simd intrinsics working across platform/arch. Possibly too into it.

cmake/ports/ffmpeg/port.cmake
```diff
@@ -259,6 +259,16 @@ if("x264" IN_LIST features)
   target_link_libraries(avcodec INTERFACE x264)
 endif()
 
+if("libopus" IN_LIST features)
```
kasperisager, line 262:
```suggestion
if("opus" IN_LIST features)
```

CMakeLists.txt
```diff
@@ -6,9 +6,11 @@ find_package(cmake-ports REQUIRED PATHS node_modules/cmake-ports)
 
 project(bare_ffmpeg C CXX)
 
+bare_arch(arch)
+
```
kasperisager, line 10:
```suggestion
```

cmake/ports/ffmpeg/port.cmake
```diff
@@ -259,6 +259,16 @@ if("x264" IN_LIST features)
...
+  find_port(libopus)
+
+  list(APPEND depends libopus)
```
kasperisager, line 265:
```suggestion
  list(APPEND depends opus)
```

cmake/ports/ffmpeg/port.cmake
```diff
@@ -259,6 +259,16 @@ if("x264" IN_LIST features)
...
+  list(APPEND args --enable-libopus)
+  list(APPEND pkg_config_path "${libopus_PREFIX}/lib/pkgconfig")
+
+  target_link_libraries(avcodec INTERFACE libopus)
```
kasperisager, line 269:
```suggestion
  target_link_libraries(avcodec INTERFACE opus)
```

kasperisager (changes requested):

cmake/ports/opus/port.cmake
```diff
@@ -0,0 +1,58 @@
...
+bare_arch(arch)
+
+if(arch MATCHES "arm64")
+  list(APPEND args
+    -DOPUS_USE_NEON=ON
+  )
+elseif(arch MATCHES "x64")
+  list(APPEND args
+    -DOPUS_X86_MAY_HAVE_SSE4_1=ON
+    -DOPUS_X86_MAY_HAVE_AVX2=ON
```
kasperisager, line 26: These enable runtime checks for SSE 4.1 and AVX2 when compiling for Intel so we can set them unconditionally, no need for check.

cmake/ports/opus/port.cmake
```diff
@@ -0,0 +1,58 @@
...
+endif()
+
+if(MSVC AND arch MATCHES "x64")
+  list(APPEND args
+    -DCMAKE_C_FLAGS=-msse4.1
+  )
+endif()
+
+declare_port(
+  "github:xiph/opus#c0eb2ca"
```
kasperisager, line 37:
```suggestion
  "github:xiph/opus@1.5.2"
```

Favor tagged releases when available, it makes it easier to identify the version and also enables shallow clones in `cmake-fetch` for faster builds.

cmake/ports/opus/port.cmake
sethvincent: Removing that arch check has the surprising outcome of this error on android-arm: `ERROR: opus not found using pkg-config`

I'm realizing that's why I got confused and thought the `-DOPUS_USE_NEON=ON` was needed because I was trying all the things that might get android-arm running (even though that's a nonsensical way to try and solve that error).

https://github.com/holepunchto/bare-ffmpeg/actions/runs/15827585515/job/44611710461#step:11:59

cmake/ports/opus/port.cmake
```diff
@@ -0,0 +1,58 @@
...
+elseif(arch MATCHES "x64")
+  list(APPEND args
+    -DOPUS_X86_MAY_HAVE_SSE4_1=ON
+    -DOPUS_X86_MAY_HAVE_AVX2=ON
+  )
+endif()
+
+if(MSVC AND arch MATCHES "x64")
+  list(APPEND args
+    -DCMAKE_C_FLAGS=-msse4.1
```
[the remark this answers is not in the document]
sethvincent: There have already been a couple attempted pull requests for this ([1](https://github.com/xiph/opus/pull/257), [2](https://github.com/xiph/opus/pull/347) on GitHub that haven't been merged.

They have a gitlab issue that seems to be the same bug here: https://gitlab.xiph.org/xiph/opus/-/issues/2366

I went ahead and used the [more recent existing pr](https://github.com/xiph/opus/pull/347) and it seems to be doing well.

cmake/ports/opus/port.cmake
kasperisager: `ON` is actually the default value for both of those options so we should be able to remove them entirely.

kasperisager (approved):

PR #47, opened 2025-06-20
Add `Symbol.dispose` methods to resource objects

kasperisager:


tony-go (commented):
This is awesome.

I wonder if we should add something in the doc? We mention the `.destroy` but what about that one?

kasperisager:
It's not yet super widely supported, we can still mention that it's available though 👍

PR #50, opened 2025-06-24
add pulseaudio & jackd support

telamon:
Enabled `libpulse` and `libjack` for linux builds

### DIFF
--- a/cmake/ports/ffmpeg/port.cmake
+++ b/cmake/ports/ffmpeg/port.cmake
@@ -88,7 +88,10 @@ elseif(LINUX)
     "--sysroot=${CMAKE_SYSROOT}"
 
     --enable-pthreads
+    --enable-libpulse
+    --enable-libjack
   )
+  target_link_libraries(avdevice INTERFACE pulse jack)
 elseif(ANDROID)
   list(APPEND args
     --target-os=android


kasperisager (changes requested):
You need to define appropriate feature flags and add ports for compiling PulseAudio and JACK from source. See https://github.com/holepunchto/bare-ffmpeg/pull/37 for an example.

telamon:
@kasperisager 

> You need to define appropriate feature flags and add ports for compiling PulseAudio and JACK from source. See #37 for an example.

On second thought.. why would we compile the sound-servers from source? We're not embedding nor cross compiling pulseaudio/jack into bare-ffmpeg; rather we're dynamic linking towards existing software that only exists on linux/desktop?

kasperisager:
No dice, it adds direct dependencies on both `libpulse.so` and `libjack.so`:

```console
$ readelf -d build/bare-ffmpeg\@1.bare 

Dynamic section at offset 0x19be0b8 contains 32 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libpulse.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libjack.so.0]
...
```

The addon must be self-contained so we must define ports to build static versions of those libraries if we want to include them.

telamon:
Alright, lemme see if i can get this sorted.

telamon:
@sethvincent It shouldn't be necessary, pipewire is slowly superseding pulseaudio but it usually ships with a `pipewire-pulse` daemon that speaks pulse protocol.

PR #54, opened 2025-06-26
Remove unused destructors

kasperisager:


lib/codec-parameters.js
```diff
@@ -6,10 +6,6 @@ module.exports = class FFmpegCodecParameters {
     this._handle = handle
   }
 
-  destroy() {
-    this._handle = null
```
kasperisager, line 10: This does nothing and was simply a placeholder from when the class implemented reference counting.

lib/stream.js
```diff
@@ -14,11 +14,6 @@ module.exports = class FFmpegStream {
     )
   }
 
-  destroy() {
-    this._codecParameters.destroy()
```
kasperisager, line 18: This is gone now so streams own no resources.

telamon (approved):

PR #55, opened 2025-06-26
add AVAudioFiFo api bindings

sethvincent:
This makes it so we can use ffmpeg's fifo api for buffering samples.

I'm wondering if it would be worth it to have a queue (probably js-side?) to handle back pressure if a producer of samples is a lot faster than the consumer. That could be added to this pr or a follow-up if needed.


binding.cc
```diff
@@ -1608,6 +1613,153 @@ bare_ffmpeg_channel_layout_from_mask(
...
+) {
+  int err;
+  int len = av_audio_fifo_drain(fifo->handle, nb_samples);
+  if (len < 0) {
+    err = js_throw_error(env, NULL, av_err2str(len));
+    assert(err == 0);
+    throw js_pending_exception;
+  }
+  return len;
+}
```
tony-go, line 1734: For consitency

```suggestion
static int
bare_ffmpeg_audio_fifo_drain(
  js_env_t *env,
  js_receiver_t,
  js_arraybuffer_span_of_t<bare_ffmpeg_audio_fifo_t, 1> fifo,
  int32_t nb_samples
) {
  int err;
 
  int len = av_audio_fifo_drain(fifo->handle, nb_samples);

  if (len < 0) {
    err = js_throw_error(env, NULL, av_err2str(len));
    assert(err == 0);
    throw js_pending_exception;
  }
 
  return len;
}
```

test/audio-fifo.js
```diff
@@ -0,0 +1,145 @@
...
+  readFrame.format = sampleFormat
+  readFrame.nbSamples = nbSamples
+  readFrame.channelLayout = channelLayout
+  readFrame.alloc()
+
+  const read = fifo.read(readFrame, nbSamples)
+  t.is(read, nbSamples, 'reads all samples')
+  t.is(fifo.size, 0)
+  t.is(fifo.space, nbSamples)
+})
```
tony-go, line 35: I would favour having two different tests one which asserts for read and one for write. The point is that in case of regression it's easier to understand how it is going.

test/audio-fifo.js
```diff
@@ -0,0 +1,145 @@
...
+
+test('AudioFifo - peek & drain', (t) => {
```
tony-go, line 37: Same for that one.

tony-go (changes requested):
Pretty neat!

We just miss the doc :)

binding.cc
```diff
@@ -1608,6 +1613,153 @@ bare_ffmpeg_channel_layout_from_mask(
...
+    nb_samples
+  );
+
+  if (fifo->handle == NULL) {
+    err = js_throw_error(env, NULL, "Failed to allocate audio FIFO");
+    assert(err == 0);
+
+    throw js_pending_exception;
+  }
+
```
kasperisager, line 1644:
```suggestion
```

We generally let OOM errors fall through. We might consider adding `assert()`s for all standard library `malloc()`s, however. V8 will catch the `ArrayBuffer` ones and terminate the process.

kasperisager (changes requested):

test/audio-fifo.js
sethvincent: 👍

test/audio-fifo.js
```diff
@@ -0,0 +1,117 @@
+const test = require('brittle')
+const ffmpeg = require('..')
+
+test('AudioFifo - write', (t) => {
+  const nbSamples = 1024
+  using fifo = createAudioFifo({ nbSamples })
```
sethvincent, line 6: Is this a valid use of `using`?

I already just learned I _do not_ want to use `using` inside the function 😏

kasperisager: Yep, that's perfectly valid!

kasperisager (approved):

PR #57, opened 2025-07-01
Fix timestamps

telamon:
- add accessors for `dts`, `pts` on `Packet`, `Frame`
- add accessors for `timeBase` on `Stream`, `Packet`, `Frame`
- add timestamp converters `packet.ptsFor(timeBase)`, `packet.dtsFor(timeBase)`, `frame.ptsFor(timeBase)`
- fix `Scaler.scale()` no longer discards packet-metadata
- add regression test for trans-coding to detect metadata loss and conversion fail


telamon:
backported into #64

PR #64, opened 2025-07-04
[WIP] Output format

telamon:


tony-go:
Hey @telamon 👋🏼 

What I propose is to split this PR into smaller chuncks, like this one: https://github.com/holepunchto/bare-ffmpeg/pull/67

lib/codec-context.js
```diff
@@ -105,8 +113,9 @@ module.exports = class FFmpegCodecContext {
     return binding.receiveCodecContextPacket(this._handle, packet._handle)
   }
 
+  /** @param {Frame|undefined} frame */
   sendFrame(frame) {
-    return binding.sendCodecContextFrame(this._handle, frame._handle)
+    return binding.sendCodecContextFrame(this._handle, frame?._handle)
```
kasperisager, line 118: We'll need to drop the `?.` syntax so we maintain support with JavaScript engines that don't support it yet, JerryScript being one.

lib/codec-parameters.js
```diff
@@ -34,17 +34,45 @@ module.exports = class FFmpegCodecParameters {
...
+      'height',
+      'bitRate',
+      'bitsPerCodedSample',
+      'bitsPerRawSample',
+      'format',
+      'sampleRate',
+      'nbChannels',
+      'channelLayout'
+    ].reduce((map, key) => ((map[key] = this[key]), map), {})
+  }
```
kasperisager, line 77: Spell these out instead and make sure to add a `__proto__` property so the class name shows up when inspected, you can find an example here: https://github.com/holepunchto/bare-url/blob/0a32f3f9330f3e3eb9216a000c57c44caf0206ea/index.js#L220-L235

lib/io-context.js
```diff
@@ -1,12 +1,35 @@
...
-      buffer.byteOffset,
-      buffer.byteLength
-    )
+  /**
+   * TODO: (help) redesign class to enable streaming IO or filehandles
+   *
+   * @param {(Buffer|number|undefined)} buffer - data | write buffer size
+   * @param {(chunk: Buffer) => void} onwrite - Write callback
+   */
+  constructor(buffer, onwrite = undefined) {
```
kasperisager, line 10: This should accept an `opts` dictionary with optional arguments such as I/O callbacks.

kasperisager (changes requested):

kasperisager:
Please remove the ad-hoc JSDoc type annotations as well before this is merged. I'm not opposed to us adding type annotations, but it has to be a concerted effort with a clear plan.

binding.cc
```diff
@@ -116,18 +165,40 @@ bare_ffmpeg_io_context_init(
...
-  uint8_t *io;
+  context->env = env;
+
+  int write_flag = 0;
+
+  if (on_write) {
+    write_flag = 1;
+    err = js_create_reference(env, *on_write, context->on_write);
+    assert(err == 0);
+  }
```
tony-go, line 176:
```suggestion
  int write_flag = 0;
  if (on_write) {
    write_flag = 1;
    err = js_create_reference(env, *on_write, context->on_write);
    assert(err == 0);
  }
```

binding.cc
```diff
@@ -116,18 +165,40 @@ bare_ffmpeg_io_context_init(
...
+  uint8_t *io = NULL;
+
+  if (size) {
     io = reinterpret_cast<uint8_t *>(av_malloc(size));
+  }
 
-    memcpy(io, &data[static_cast<size_t>(offset)], size);
+  if (data) {
+    memcpy(io, &data.value()[static_cast<size_t>(offset)], size);
   }
```
tony-go, line 188: I wonder if it could not work like that.

```cc
  uint8_t *io = NULL;
  size_t size = static_cast<size_t>(len);
  if (size) {
    io = reinterpret_cast<uint8_t *>(av_malloc(size));
    memcpy(io, &data.value()[static_cast<size_t>(offset)], size);
  }
```

binding.cc
```diff
@@ -404,16 +493,174 @@ bare_ffmpeg_format_context_read_frame(
...
+  std::optional<js_arraybuffer_span_of_t<bare_ffmpeg_dictionary_t, 1>> muxer_options
+) {
+  int err;
+
+  if (!muxer_options) {
+    err = avformat_write_header(context->handle, NULL);
+  } else {
+    auto dict = *muxer_options;
+
+    err = avformat_write_header(context->handle, &dict->handle);
```
tony-go, line 521: Maybe we would like to assert `err`?

tony-go, line 521:
```suggestion
    auto dict = *muxer_options;
    err = avformat_write_header(context->handle, &dict->handle);
```

binding.cc
```diff
@@ -404,16 +493,174 @@ bare_ffmpeg_format_context_read_frame(
...
+
+    const AVDictionaryEntry *option = NULL;
+
+    while ((option = av_dict_iterate(dict->handle, option))) {
+      av_log(context->handle, AV_LOG_WARNING, "Ignored option key='%s' value='%s'\n", option->key, option->value);
+    }
```
tony-go, line 527:
```suggestion
    const AVDictionaryEntry *option = NULL;
    while ((option = av_dict_iterate(dict->handle, option))) {
      av_log(context->handle, AV_LOG_WARNING, "Ignored option key='%s' value='%s'\n", option->key, option->value);
    }
```

binding.cc
```diff
@@ -404,16 +493,174 @@ bare_ffmpeg_format_context_read_frame(
...
+  js_arraybuffer_span_of_t<bare_ffmpeg_format_context_t, 1> context
+) {
+  int err = av_write_trailer(context->handle);
+
+  if (err < 0) {
+    err = js_throw_error(env, NULL, av_err2str(err));
+    assert(err == 0);
+
+    throw js_pending_exception;
+  }
```
tony-go, line 586:
```suggestion
  int err = av_write_trailer(context->handle);
  if (err < 0) {
    err = js_throw_error(env, NULL, av_err2str(err));
    assert(err == 0);

    throw js_pending_exception;
  }
```

test/format-context-output.js
```diff
@@ -0,0 +1,404 @@
...
+        } catch (error) {
+          t.comment('resource dispose failed for', r)
+          throw error
+        }
+      }
+
+      cleaning = false
+    }
+  }
+}
```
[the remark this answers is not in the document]
telamon: agreed this tool is misplaced.
I'm thinking to just stash it into another module or maybe PR an ownership assertion feature to `brittle`?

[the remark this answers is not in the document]
telamon: Just to clarify i've stumbled on instances of resources that are either not disposed or disposed twice in other parts of the api. so am proposing to ban keyword `using` from tests and assert ownership. let me readdress this in an issue.

binding.cc
telamon: It does not sadly / segfaults, i tried skipping the allocation at initially but ffmpeg expects the first buffer to be fully initialized.

tony-go: Yeah I did not skiped the allocation, I wonder just if there a `size` without `data`?

Plus, what if we have `data` and size `zero`? We don't do `malloc` ? (really question, I could I ve miss a part)

binding.cc
```diff
@@ -404,16 +493,171 @@ bare_ffmpeg_format_context_read_frame(
...
+
+  // TODO: expose as packet.convertTimebase(Rational)
+  if (!bad_timebase(srcTime) && !bad_timebase(dstTime)) {
+    av_packet_rescale_ts(packet->handle, srcTime, dstTime);
+    packet->handle->time_base = dstTime;
+  }
+#endif
+
+  // remove after identifying stream
+  // printf("ctx_write_packet stream[%p]=(%i / %i) packet: stream=%i tb=(%i / %i) pts=%zi duration=%zi\n", stream, stream->time_base.num, stream->time_base.den, packet->handle->stream_index, packet->handle->time_base.num, packet->handle->time_base.den, packet->handle->pts, packet->handle->duration);
```
tony-go, line 559: Let's remove this.

binding.cc
```diff
@@ -791,6 +1062,16 @@ bare_ffmpeg_codec_context_send_packet(
 ) {
   int err;
 
+#ifdef AUTO_TIMEBASE
+  if (
+    !bad_timebase(packet->handle->time_base) &&
+    bad_timebase(context->handle->pkt_timebase)
+  ) {
+    // Auto init decoder time_base
+    context->handle->pkt_timebase = packet->handle->time_base;
```
tony-go, line 1071:
```suggestion
    context->handle->pkt_timebase = packet->handle->time_base;
```

test/packet.js
```diff
@@ -9,11 +9,15 @@ test('packet should expose a buffer getter', (t) => {
...
+test('packet should expose a streamIndex accessor', (t) => {
   const packet = new ffmpeg.Packet()
   fillPacket(packet)
 
   t.ok(typeof packet.streamIndex == 'number')
+
+  packet.streamIndex = 9
+
+  t.is(packet.streamIndex, 9)
 })
```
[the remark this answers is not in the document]
tony-go: I usually test them separately:
- setter:  I just test that the function is exposed and does not crash + validation
- getter: I just test that the function is exposed and does not crash + validation + it returns the proper value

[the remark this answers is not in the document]
telamon: I assumed that was the intention -  but that not what the separate tests do:
```js
test('frame expose a getter for width', (t) => {
  const fr = new ffmpeg.Frame()
  // NOTE: Invokes the setter or if setter is not exposed then assign 200 as regular JS property.
  fr.width = 200 

  t.ok(fr.width == 200) // possibly tests js-prop
})

test('frame expose a setter for height', (t) => {
  const fr = new ffmpeg.Frame()
  t.execution(() => { // redundant, if execution fails it will do so in previous test
    fr.height = 200 // possibly sets js prop
  })
})
```

Where as the accessor tests:

```js
test('frame expose a accessor for pts', (t) => {
  const fr = new ffmpeg.Frame()

  t.is(fr.pts, -1) // invokes getter, validates initial value set by ffmpeg

  fr.pts = 0 // invoke setter

  t.is(fr.pts, 0) // validate set value or fail if setter not implemented.
})
```

binding.cc
```diff
@@ -791,6 +1062,16 @@ bare_ffmpeg_codec_context_send_packet(
 ) {
   int err;
 
+#ifdef AUTO_TIMEBASE
+  if (
+    !bad_timebase(packet->handle->time_base) &&
+    bad_timebase(context->handle->pkt_timebase)
+  ) {
```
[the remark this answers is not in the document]
telamon: I would too but JS does not support 64bit integers so all timestamp & timebase calculations must be done on native side

binding.cc
telamon: Sorry this was not clearly formulated, there is something weird in ffmpeg,
symptom is:
```C
AVStream* stream = avformat_create_stream(format_context);
int idx = stream->index;

// later during write_frame
assert(stream == format_context->streams[idx]); // fails
```
I saw a set of completely different pointers to what looked like a pair of default initialized streams.
Not at all the streams i created and configured.

binding.cc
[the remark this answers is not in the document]
telamon: Experimental feature discarded, in favor of public API
- added timeBase accessors for `Stream`, `Packet`, `FormatContext`, `CodecContext`, `CodecParameters`
- add `Packet.rescale(otherTimebase)`

mafintosh:
Sounds like all issues resolved, clicking the button

PR #67, opened 2025-07-10
feat(stream): enhance api

tony-go:
Extracted from https://github.com/holepunchto/bare-ffmpeg/pull/64/files


tony-go:
I pushed test directly on #64 !

PR #70, opened 2025-07-16
Add visual mental map to the readme

RaisinTen:
This adds the visual mental map of the bare-ffmpeg APIs created by @telamon to the README. It gives an overview of all the FFmpeg APIs and how they relate to one another. This can be modified by dragging and dropping it into the <https://www.drawio.com> editor.


telamon:
Happy you liked it :-) the map is no-where complete; source-code is always the best source of truth / docs.
But having said that, not sure it should be presented in the readme but a `contrib/cheatsheet.png` wouldn't hurt.

RaisinTen:
@kasperisager suggested using a mermaid version instead of committing a blob into the repo, so we can easily maintain the diagram as we iterate the API. It won't be as visually pretty, but it'll be code that we can modify and maintain.

Once that's ready, I can move it under `contrib`. :)

RaisinTen:
Closing this as discussed with Tony because mermaid diagrams gets cluttered fast and are not really human friendly

PR #75, opened 2025-07-29
Increase `FormatContext` coverage

yassernasc:
Those are basic informative API's, 

The `duration` value is available at file headers depending on the file format, so we have this easy access option. After testing with the mp3 sample it returned a negative number that exceeded the Number.MIN_SAFE_INTEGER, so maybe that's a footgun, FFmpeg seems to have funky default values.


test/input-format.js
```diff
@@ -2,5 +2,20 @@ const test = require('brittle')
...
+
+  t.is(input.name, 'aiff')
+  t.is(input.longName, 'Audio IFF')
+})
+
+test('OutputFormat info', (t) => {
+  const output = new ffmpeg.OutputFormat('webm')
+
+  t.is(output.name, 'webm')
+  t.is(output.longName, 'WebM')
```
tony-go, line 20: Maybe we could like to add `jpeg` too?

test/format-context.js
```diff
@@ -63,10 +58,41 @@ test('InputFormatContext.getBestStream should return a null if no stream is foun
   t.is(bestStream, null)
 })
 
+test('duration', (t) => {
```
tony-go, line 61:
```suggestion
test('InputFormatContext should expose a duration getter', (t) => {
```

test/format-context.js
```diff
@@ -63,10 +58,41 @@ test('InputFormatContext.getBestStream should return a null if no stream is foun
...
+  using imageFormatContext = getInputFormatContext('jpeg')
+  using aiffFormatContext = getInputFormatContext('aiff')
+  using mp3FormatContext = getInputFormatContext('mp3')
+
+  t.is(imageFormatContext.duration, 0)
+  t.is(aiffFormatContext.duration, 2936625)
+  t.is(mp3FormatContext.duration, 0)
+})
+
+test('inputFormat getter', (t) => {
```
tony-go, line 71:
```suggestion
test('InputFormatContext should expose a inputFormat getter', (t) => {
```

test/format-context.js
```diff
@@ -63,10 +58,41 @@ test('InputFormatContext.getBestStream should return a null if no stream is foun
   t.is(bestStream, null)
 })
 
+test('duration', (t) => {
+  using imageFormatContext = getInputFormatContext('jpeg')
+  using aiffFormatContext = getInputFormatContext('aiff')
+  using mp3FormatContext = getInputFormatContext('mp3')
+
+  t.is(imageFormatContext.duration, 0)
+  t.is(aiffFormatContext.duration, 2936625)
```
tony-go, line 67: This is probably related to the comment you let in the description.

We should investigate on that.

binding.cc
```diff
@@ -468,6 +504,38 @@ bare_ffmpeg_format_context_create_stream(
...
 
+static int64_t
+bare_ffmpeg_format_context_get_duration(
+  js_env_t *env,
+  js_receiver_t,
+  js_arraybuffer_span_of_t<bare_ffmpeg_format_context_t, 1> context
+) {
+  int64_t duration = context->handle->duration;
+
+  if (duration == -0x8000000000000000) return 0;
```
tony-go, line 515: I wonder if it could be a use case for big ints... 🤔

yassernasc:
closing by now as non-priority

PR #77, opened 2025-07-31
Cleanup `ffmpeg.Rational`

telamon:
- moved getter `rational.q2d` to function  `rational.toNumber()`
- add js-impl of `av_rescale_q()` as `Rational.rescaleQ(n, aBase, bBase)`
- updated typedefs
- registered `test/rational.js` in test-runner
- updated `README.md`


tony-go (commented):
Also we would like to have the doc updated. 🙌🏼

lib/rational.js
```diff
@@ -11,22 +11,29 @@ module.exports = class FFmpegRational {
...
-    if (this.q2d === 0) return false
+    if (this.toNumber() === 0) return false
 
     return true
   }
 
   get uninitialized() {
+    // common initial value for AVRational(0, 1)
     return this.numerator === 0 && this.denominator === 1
   }
```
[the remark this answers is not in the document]
telamon: you mean https://github.com/holepunchto/bare-ffmpeg/pull/77/files#diff-ee8ec00fc34847b0434f02f5bfd5d8993d709f438211eab0032a136c79ef211fR11-R12
?

tony-go (approved):

lib/rational.js
```diff
@@ -7,26 +7,34 @@ module.exports = class FFmpegRational {
...
     const view = new Int32Array(binding.rationalD2Q(num))
     return new FFmpegRational(view[0], view[1])
   }
+
+  static rescaleQ(n, src, dst) {
+    return (
+      (n * (src.numerator / src.denominator)) /
+      (dst.numerator / dst.denominator)
+    )
+  }
```
tony-go, line 39: The PR looks good to me! Ready to merge.

But I wonder if we would not let that kind of code leave into the userland.

I ask because the spirit of that biding is to reproduce the low-level API as much as we can. So I just wonder if we don't betray the spirit.

More a question than a strong opinion.

Maybe @kasperisager has an touhoughts.

yassernasc: I share the concern, I'd recommend using https://ffmpeg.org/doxygen/trunk/group__lavu__math.html#gaf02994a8bbeaa91d4757df179cbe567f even if the practical results are the same

kasperisager: Reimplementing the FFmpeg API in JavaScript makes good sense in this particular case as the overhead of calling into native for the sake of some mathematical operations will be unreasonably high.

kasperisager (approved):

PR #84, opened 2025-08-07
Expose an API to get the Frame data at a plane

RaisinTen:
This is needed for audio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,12 @@ Destroys the `Frame` and frees all associated resources. Automatically called wh
 
 **Returns**: `void`
 
+##### `Frame.data(plane)`
+
+Gets the frame data buffer for the given plane.
+
+**Returns**: `Buffer`
+
 ### `Packet`
 
 This structure stores compressed data. It is typically exported by demuxers and then passed as input to decoders, or received as output from encoders and then passed to muxers.
--- a/README.md
+++ b/README.md
@@ -480,6 +480,10 @@ Destroys the `Frame` and frees all associated resources. Automatically called wh
 
 Gets the frame data buffer for the given plane.
 
+Parameters:
+
+- `plane` (`number`): The plane index
+
 **Returns**: `Buffer`
 
 ### `Packet`
--- a/binding.cc
+++ b/binding.cc
@@ -1444,7 +1444,7 @@ bare_ffmpeg_frame_get_data(
     throw js_pending_exception;
   }
 
-  uint8_t* src = frame->handle->data[plane];
+  uint8_t *src = frame->handle->data[plane];
   int linesize = frame->handle->linesize[plane];
   size_t copy_size = 0;
 
@@ -1459,7 +1459,7 @@ bare_ffmpeg_frame_get_data(
     // Audio frame
     // Assume planar if each plane exists separately.
     if (av_sample_fmt_is_planar((AVSampleFormat) frame->handle->format)) {
-      int bytes_per_sample = av_get_bytes_per_sample((AVSampleFormat)frame->handle->format);
+      int bytes_per_sample = av_get_bytes_per_sample((AVSampleFormat) frame->handle->format);
       copy_size = frame->handle->nb_samples * bytes_per_sample;
     } else {
       // Packed audio, everything is in data[0]
--- a/binding.cc
+++ b/binding.cc
@@ -1429,6 +1429,70 @@ bare_ffmpeg_frame_destroy(
 }
 
 static int32_t
+bare_ffmpeg_frame_get_data(
+  js_env_t *env,
+  js_receiver_t,
+  js_arraybuffer_span_of_t<bare_ffmpeg_frame_t, 1> frame,
+  int32_t plane
+) {
+  int err;
+
+  if (plane < 0 || plane >= AV_NUM_DATA_POINTERS || !frame->handle->data[plane]) {
+    err = js_throw_errorf(env, NULL, "Invalid or empty plane %d", plane);
+    assert(err == 0);
+
+    throw js_pending_exception;
+  }
+
+  uint8_t* src = frame->handle->data[plane];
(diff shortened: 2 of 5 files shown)


tony-go (commented):
@RaisinTen Could you please add unit tests to asserts different cases we have here.

RaisinTen:
@sethvincent said:
> can confirm that `frame.data` does have the audio data and `av_samples_fill_arrays` is being used incorrectly. exposing a function like `FFFmpegFrame.data(plane)` will be pretty straightforward. i can do that in pst morning if nobody beats me to it

I opened this PR, so that we can test playing the raw audio data in https://github.com/holepunchto/call-doctor/pull/16 by applying this diff on this line https://github.com/holepunchto/call-doctor/pull/16/files#diff-d2e1619fcd0984e5c9ac52481cb40f9c39ce4c9e9b4bc01ea3878aeaba7d1910R188
```diff
-        const buffer = samples.data
+        const buffer = Buffer.from(decodedAudioFrame.data(0))
```

Also, cc @telamon .

RaisinTen:
I did some digging and my understanding is that the bare-ffmpeg code is correct. It is expected that `samples.data` will be 0-filled before resampling because the data gets filled only by the `resampler.convert(decodedAudioFrame, output)` step.

Here's a breakdown of how the `samples.data` buffer gets filled.
```js
const audioDecoder = bestAudioStream.decoder()
const decodedAudioFrame = new ffmpeg.Frame()
const resampler = new ffmpeg.Resampler(
  bestAudioStream.codecParameters.sampleRate,
  bestAudioStream.codecParameters.channelLayout,
  bestAudioStream.codecParameters.format,
  bestAudioStream.codecParameters.sampleRate,
  ffmpeg.constants.channelLayouts.STEREO,
  ffmpeg.constants.sampleFormats.S16
)
while (audioDecoder.receiveFrame(decodedAudioFrame)) {
  // Now, decodedAudioFrame is populated with audio data in the received format.

  // Creates an empty output frame.
  const output = new ffmpeg.Frame()
  output.channelLayout = ffmpeg.constants.channelLayouts.STEREO
  output.format = ffmpeg.constants.sampleFormats.S16
  output.nbSamples = decodedAudioFrame.nbSamples
  output.sampleRate = bestAudioStream.codecParameters.sampleRate

  // Now, samples.data is a 0-filled buffer of the calculated size.
  const samples = new ffmpeg.Samples('S16', 2, output.nbSamples)

  // This populates the empty output frame by doing:
  // output._handle.data[0] = samples.data
  // Refs: https://ffmpeg.org/doxygen/2.4/samplefmt_8c_source.html#l00169
  samples.fill(output)

  // This resamples the incoming audio data frame and writes it to the output frame at output._handle.data which holds samples.data at index 0.
  // Refs: https://github.com/holepunchto/bare-ffmpeg/blob/e21b24aab6ae8f94b6cedf458646c8eb3759b169/binding.cc#L2345
  resampler.convert(decodedAudioFrame, output)

  // Now, samples.data holds the audio data in the desired format and can be played.
  play(samples.data)
}
```

kasperisager:
That's correct, yes 👍

telamon:
But then the bug you had was just an order-of-operations issue,
will this produce samples?
```js
const audioDecoder = bestAudioStream.decoder()
const decodedAudioFrame = new ffmpeg.Frame()
const samples = new ffmpeg.Samples('S16', 2, output.nbSamples)
samples.fill(decodedAudioFrame)

while (audioDecoder.receiveFrame(decodedAudioFrame)) {
  play(samples.data) // non-zero
}

**Comment by RaisinTen** [comment]
@telamon I tried it and the `samples.data` is 0-filled.

Note that `nbSamples` is unknown until the first audio frame is fetched, so I hardcoded that to `3800` locally just for testing. In your example, when you run `samples.fill(decodedAudioFrame)`, it would do:
```js
decodedAudioFrame._handle.data[0] = samples.data
```
So index 0 will be a pointer to a 0-filled buffer because that's what `samples.data` is originally.

And then when `audioDecoder.receiveFrame(decodedAudioFrame)` gets called, the first thing this does is call
```
av_frame_unref(frame);
```
in https://ffmpeg.org/doxygen/6.1/avcodec_8c_source.html#l00715 which means that it unreferences all the buffers referenced by frame and reset the frame fields, so the connection with `samples.data` is gone and so `samples.data` will remain unaffected by further operations and stay a 0-filled buffer.

kasperisager:
That also makes sense; `Samples` is used, as in your previous example, for zero-copy access to resampled audio frames.

telamon:
> And then when `audioDecoder.receiveFrame(decodedAudioFrame)` gets called, the first thing this does is call
> 
> ```
> av_frame_unref(frame);
> ```
> 
> in https://ffmpeg.org/doxygen/6.1/avcodec_8c_source.html#l00715 which means that it unreferences all the buffers referenced by frame and reset the frame fields, so the connection with `samples.data` is gone and so `samples.data` will remain unaffected by further operations and stay a 0-filled buffer.

@RaisinTen aha! nice sleuthing, but that breaks zero-copy yes and forces use of a resampler / copier.

@kasperisager do you know why we're doing `av_frame_unref()` in `codecContext.receiveFrame()` ?

telamon:
@kasperisager Gotcha, would it make sense to expose `frame.getChannelData(channel_idx)` and return the buffer?

kasperisager:
What's the exact use case here that can't be solved by passing through the `Resampler` and `Samples` API?

kasperisager:
Are either of those actual issues when using `Resampler` and `Samples` though? Unless they are then I don't think we should add additional and quite complex API surface. It'll come back to bite us.

RaisinTen:
As @tony-go mentioned on keet, I'm closing this and opening an issue instead https://github.com/holepunchto/bare-ffmpeg/issues/97

PR #88, opened 2025-08-08
fix(codec-parameters): dedup format getter and rename setter

tony-go:


kasperisager (approved):
There's a bit of a stutter in the binding methods for the codec parameters, e.g. `setCodecParametersCodecFormat()` should be just `setCodecParametersFormat()`, but that can be fixed later 👌

tony-go:
> There's a bit of a stutter in the binding methods for the codec parameters, e.g. `setCodecParametersCodecFormat()` should be just `setCodecParametersFormat()`, but that can be fixed later 👌

This is just what I post on Keet ahah! Next PR is for that!

PR #91, opened 2025-08-08
feat(codec-parameters): setter `channelLayout`

telamon:


lib/codec.js
```diff
@@ -27,7 +27,9 @@ module.exports = class FFmpegCodec {
   }
 
   get name() {
-    return binding.getCodecNameByID(this._id)
+    if (Number.isInteger(this._id)) {
+      return binding.getCodecNameByID(this._id)
+    }
```
tony-go, line 32: WDYT about adding a unit test!

If you are on a rush, lets add a todo and I'll do it into a subsequent PR

telamon: lemme add it, it was a tiny bugfix that appeared when trying to `console.log(codec)`

tony-go (approved):

PR #95, opened 2025-08-12
read/seek for streaming iocontext

sethvincent:
This adds onread/onseek callbacks for IOContext.


test.js
```diff
@@ -14,3 +14,4 @@ require('./test/resampler')
 require('./test/stream')
 require('./test/log')
 require('./test/rational')
+require('./test/io-context-streaming')
```
tony-go, line 17: I think that we should have rather a

```suggestion
require('./test/io-context')
```

test/io-context-streaming.js
```diff
@@ -0,0 +1,121 @@
+const test = require('brittle')
+const ffmpeg = require('..')
+
+const { mediaTypes } = ffmpeg.constants
+
+test('IOContext streaming webm without onseek', (t) => {
```
tony-go, line 6:
```suggestion
test('IOContext streaming webm with onread', (t) => {
```

It sounds more like what I see.

test/io-context-streaming.js
```diff
@@ -0,0 +1,121 @@
...
+    if (mediaType === mediaTypes.VIDEO) {
+      video.push(packet.data)
+    } else if (mediaType === mediaTypes.AUDIO) {
+      audio.push(packet.data)
+    }
+
+    packet.unref()
+  }
+
+  for (const { decoder } of streams) decoder?.destroy()
```
tony-go, line 53: That part seems to be generic and not related to the test per se. What do you thuink about offloading it into a `readFrame` function that will return a `video` and `audio` buffer?

The idea is that we can reduce the noise between the "when" and the "then".

lib/io-context.js
```diff
@@ -17,14 +17,38 @@ module.exports = class FFmpegIOContext {
...
+          out.byteOffset || 0,
+          (out.byteOffset || 0) + out.byteLength
+        )
+      }
+    }
+
+    let onseek
+    if (opts.onseek) {
+      onseek = (newOffset) => {
+        opts.onseek(Number(newOffset))
```
tony-go, line 40: Why do we need to wrap `newOffset` into `Number` here?

lib/io-context.js
```diff
@@ -17,14 +17,38 @@ module.exports = class FFmpegIOContext {
...
 
-    this._handle = binding.initIOContext(buffer, offset, len, onwrite)
+    let onread
+    if (opts.onread) {
+      onread = (requested) => {
+        const out = opts.onread(requested)
+        return out.buffer.slice(
+          out.byteOffset || 0,
+          (out.byteOffset || 0) + out.byteLength
+        )
```
tony-go, line 33:
```suggestion
        return out.buffer.slice(
          out.byteOffset,
          out.byteOffset + out.byteLength
        )
```

test/io-context-streaming.js
```diff
@@ -0,0 +1,121 @@
...
+    onread: (len) => {
+      if (offset >= data.length) {
+        return Buffer.alloc(0)
+      }
+
+      const n = Math.min(len, data.length - offset)
+      const chunk = data.subarray(offset, offset + n)
+      offset += n
+
+      return chunk
```
tony-go, line 22: Why do we return a Buffer ? It should not be simply a number?

```c
int (*read_packet)(void *opaque, uint8_t *buf, int buf_size),
```

sethvincent: that's the signature for io_context_read_packet in binding.cc https://github.com/holepunchto/bare-ffmpeg/pull/95/files#diff-006b4a5fdf59ef5d26a069195474bba89508b525dced6cee1fa4dc6869e4e578R153

this returns the data inside that function so it can do its thing

lib/io-context.js
sethvincent: i've found that the first few return values here can have an undefined byteOffset, but yeah i should see if i can fix that earlier in the call stack.

test/io-context-streaming.js
sethvincent: heck yeah! i made a helper function in this file for now but could move it elsewhere if needed.

test/io-context-streaming.js
tony-go: yeah but we have to return it from js side?

binding.cc
```diff
@@ -33,12 +33,18 @@ extern "C" {
 }
 
 using bare_ffmpeg_io_context_on_write_cb_t = js_function_t<void, js_arraybuffer_t>;
+using bare_ffmpeg_io_context_on_read_cb_t = js_function_t<js_arraybuffer_t, int32_t>;
```
sethvincent, line 36: the last change i'm considering is making this a typedarray span so it doesn't involve a copy. could optionally make it a param that's passed from the c++ side.

@tony-go is this what you were getting at with the question about returning the arraybuffer from the onread js callback?

tony-go: @sethvincent  Not at all. I think that I misunderstood the purpose of the function in the first place.

lib/io-context.js
```diff
@@ -17,14 +17,35 @@ module.exports = class FFmpegIOContext {
...
         opts.onwrite(Buffer.from(chunk))
       }
     }
 
-    this._handle = binding.initIOContext(buffer, offset, len, onwrite)
+    let onread
+    if (opts.onread) {
+      onread = (requested) => {
+        const out = opts.onread(requested)
+        return out.buffer.slice(out.byteOffset, out.byteOffset + out.byteLength)
```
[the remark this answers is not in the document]
sethvincent: I just pushed some changes that should work! Goal is to provide a typedarray that wraps the `uint8_t *buf` from c++ to js that the onread callback can write to it.

Not sure what is meant by this jstl error:

```
/Users/runner/work/bare-ffmpeg/bare-ffmpeg/build/_deps/github+holepunchto+libjstl-src/include/jstl.h:3632:1: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'options'
 3632 | js_call_function(js_env_t *env, const js_function_t<R, A...> &function, const A &...args, R &result) {
      | ^
```

binding.cc
```diff
@@ -143,14 +149,82 @@ io_context_write_packet(void *opaque, const uint8_t *chunk, int len) {
...
+  auto context = reinterpret_cast<bare_ffmpeg_io_context_t *>(opaque);
+  auto env = context->env;
+
+  bare_ffmpeg_io_context_on_read_cb_t callback;
+  err = js_get_reference_value(env, context->on_read, callback);
+  assert(err == 0);
+
+  js_arraybuffer_t arraybuffer;
+  err = js_create_external_arraybuffer(env, buf, static_cast<size_t>(buf_size), arraybuffer);
+  assert(err == 0);
```
tony-go, line 165: Really cooler like that 😎 !

tony-go (approved):
Code is way more readable with last tweaks :)

kasperisager (changes requested):

binding.cc
sethvincent: yeah! this is much better than the previous version 😅

binding.cc
```diff
@@ -143,14 +149,82 @@ io_context_write_packet(void *opaque, const uint8_t *chunk, int len) {
...
+
+  js_typedarray_t<uint8_t> uint8array;
+  err = js_create_typedarray(env, static_cast<size_t>(buf_size), arraybuffer, uint8array);
+  assert(err == 0);
+
+  int32_t result;
+  err = js_call_function<js_type_options_t{}, int32_t, js_typedarray_t<uint8_t>, int32_t>(env, callback, uint8array, buf_size, result);
+  assert(err == 0);
+
```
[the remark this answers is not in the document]
sethvincent: ah, thanks!

PR #96, opened 2025-08-12
feat(dictionary): add iterator

### LINKED ISSUE #86: Add `Symbol.iterator` for `Dictionary`
Most operations that accept an `ffmpeg.Dictionary` clears out all consumed key/values and leaves only the unsupported key/values.
https://github.com/holepunchto/bare-ffmpeg/blob/e21b24aab6ae8f94b6cedf458646c8eb3759b169/binding.cc#L505-L511

In order to easier catch invalid `Symbol.iterator` could be implemented:

```js
class Dictionary {
  // ...

  [Symbol.iterator] () {
    return binding.iterateDictionary(this._handle) // => [[key1, value1], [key2, value2], ...]
  }
}
```

Also there is a unique behavior / bug in `format-context.js`
https://github.com/holepunchto/bare-ffmpeg/blob/e21b24aab6ae8f94b6cedf458646c8eb3759b169/lib/format-context.js#L106

The `options.destroy()` should only be called if the options were created by the constructor automatically. I cannot find any other instance of resources ownership being adopted this way so assuming it's a bug.

tony-go:
Fix #86

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -1,5 +1,6 @@
 #include <assert.h>
 #include <cstdint>
+#include <optional>
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>
@@ -2498,14 +2499,15 @@ bare_ffmpeg_dictionary_set_entry(
   assert(err == 0);
 }
 
-static std::tuple<const char *, const char *>
+static std::optional<std::tuple<const char *, const char *>>
 bare_ffmpeg_dictionary_yield_key_value(
   js_env_t *env,
   js_receiver_t,
   js_arraybuffer_span_of_t<bare_ffmpeg_dictionary_t, 1> dict
 ) {
   dict->prev = av_dict_iterate(dict->handle, dict->prev);
-  return dict->prev != nullptr ? std::tuple{dict->prev->key, dict->prev->value} : std::tuple{"", ""};
+  if (dict->prev != nullptr) return std::tuple{dict->prev->key, dict->prev->value};
+  else return std::nullopt;
 }
 
 static std::optional<std::string>
--- a/binding.cc
+++ b/binding.cc
@@ -87,6 +87,7 @@ typedef struct {
 
 typedef struct {
   struct AVDictionary *handle;
+  const struct AVDictionaryEntry *prev;
 } bare_ffmpeg_dictionary_t;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


kasperisager (changes requested):
The iterator state must be independent of the dictionary handle. The simplest approach is to simply return an array of all key-value pairs from the binding.

tony-go:
> The iterator state must be independent of the dictionary handle.

Could you explain why ? 😅

kasperisager:
Having multiple concurrent iterators will be borked otherwise, as they'll all be advancing the same underlying state.

kasperisager:
Collecting the key-value pairs into an array up-front rather than incrementally traversing the list pointers would also solve the issue of modifying the dictionary during iteration.

binding.cc
```diff
@@ -3058,6 +3071,7 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   V("destroyDictionary", bare_ffmpeg_dictionary_destroy)
   V("getDictionaryEntry", bare_ffmpeg_dictionary_get_entry)
   V("setDictionaryEntry", bare_ffmpeg_dictionary_set_entry)
+  V("getDictionnaryEntries", bare_ffmpeg_dictionary_get_entries)
```
kasperisager, line 3074:
```suggestion
  V("getDictionaryEntries", bare_ffmpeg_dictionary_get_entries)
```

lib/dictionary.js
```diff
@@ -22,6 +22,16 @@ module.exports = class FFmpegDictionary {
     binding.setDictionaryEntry(this._handle, key, value)
   }
 
+  entries() {
+    return binding.getDictionnaryEntries(this._handle)
```
kasperisager, line 26:
```suggestion
    return binding.getDictionaryEntries(this._handle)
```

lib/dictionary.js
```diff
@@ -22,6 +22,16 @@ module.exports = class FFmpegDictionary {
...
+  }
+
+  *[Symbol.iterator]() {
+    for (const entry of this.entries()) {
+      yield entry
+    }
```
kasperisager, line 32:
```suggestion
    yield* this.entries()
```

lib/dictionary.js
```diff
@@ -22,6 +22,16 @@ module.exports = class FFmpegDictionary {
...
+  }
```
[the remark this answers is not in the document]
kasperisager: It's a nice convenience, similar to how you can do `for (const entry of map)` instead of `for (const entry of map.entries())`.

PR #98, opened 2025-08-15
Disable GPL code

kasperisager:


tony-go:
Just to see if we have other issues than that one :)

tony-go (approved):
Lessgo!

I noticed that we had a [flaky test](https://github.com/holepunchto/bare-ffmpeg/actions/runs/17071613875/job/48401784644), so I created a specific issue as it is not related to this PR. (#103)

PR #106, opened 2025-08-31
reworked `IOContext` callbacks

telamon:
- write_callback;  moved typedarray view from js wrapper to native
- read/seek: removed `context.offset` member as belongs to user IO implementation.
- seek: exposed seek-operation `whence` as a string.
- seek: returning `-1` signals that operation is not supported or not implemented.

@kasperisager just to be sure - i confirmed that the callbacks run on js-stack; but do we also need to surround the calls with open/close handle-scope?

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -222,11 +222,11 @@ bare_ffmpeg__on_io_context_seek(void *opaque, int64_t offset, int whence) {
   err = js_get_reference_value(env, context->on_seek, callback);
   assert(err == 0);
 
-  err = js_call_function(env, callback, offset, whence_str, result);
-  if (err < 0) return AVERROR(EIO);
+  err = js_call_function<js_type_options_t{}, int64_t, int64_t, std::string>(env, callback, offset, whence_str, result);
+  if (err < 0) return AVERROR(EIO); // read-error
 
   if (result == -1) {
-    return AVERROR(ENOSYS); // evals to 0??
+    return AVERROR(ENOSYS); // seek-op not supported by IO
   }
 
   return result;
--- a/binding.cc
+++ b/binding.cc
@@ -33,9 +33,9 @@ extern "C" {
 #include <libswscale/swscale.h>
 }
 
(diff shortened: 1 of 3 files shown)


kasperisager:
Scopes are only needed if the JavaScript execution stack is empty, i.e. if the callbacks aren't invoked as a result of already executing JavaScript. I'm fairly sure that the FFmpeg I/O callbacks are only invoked when calling into FFmpeg APIs so there will always be an existing handle scope open.

test/io-context.js
```diff
@@ -59,8 +59,20 @@ test('IOContext streaming mp4 with onseek', (t) => {
       offset += bytesToRead
       return bytesToRead
     },
-    onseek: (newOffset) => {
-      offset = Math.max(0, Math.min(newOffset, data.length))
+
+    onseek: (o, whence) => {
```
tony-go, line 63: I created an issue for improving coverage of this namespace #107

binding.cc
```diff
@@ -146,7 +144,12 @@ bare_ffmpeg__on_io_context_write(void *opaque, const uint8_t *buf, int len) {
   err = js_create_arraybuffer(env, buf, static_cast<size_t>(len), data);
   assert(err == 0);
 
-  err = js_call_function(env, callback, data);
+  js_typedarray_t<uint8_t> view;
+  err = js_create_typedarray(env, static_cast<size_t>(len), data, view);
+  assert(err == 0);
+
+
+  err = js_call_function(env, callback, view);
```
tony-go, line 152: We passed a view in this way we avoid copy? I am not sure to understand why.

I think that I should refer to

> write_callback; moved typedarray view from js wrapper to native


@telamon

telamon: In native we usually just use `js_arraybuffer_t` because they're more practical, but exporting that to JS gives you an `ArrayBuffer` which can't be read/written directly without a view such as:

```js
const buffer = Buffer.from(arrayBuffer)
```
I just moved this line :arrow_up:  to native instead since the other two callbacks were written that way

PR #113, opened 2025-09-03
fix(io-context): handle onerror and onseek error propagation

tony-go:


binding.cc
```diff
@@ -373,6 +373,10 @@ bare_ffmpeg_format_context_open_input_with_io(
   if (err < 0) {
     avformat_free_context(context->handle);
 
+    bool is_exception_pending;
+    js_is_exception_pending(env, &is_exception_pending);
+    if (is_exception_pending) throw js_pending_exception;
```
kasperisager, line 378:
```suggestion
    err = js_is_exception_pending(env, &is_exception_pending);
    assert(err == 0);

    if (is_exception_pending) throw js_pending_exception;
```

kasperisager (changes requested):

kasperisager (approved):

PR #119, opened 2025-09-09
expose `packet.sideData`

telamon:
Added quick and dirt `packet.sideData` serialization trying to fix:
https://github.com/holepunchto/vcr-core/issues/31

Suggestions on better encoding welcome, these functions are in a hot-path.

---

Additional edits:
- removed return statements from setters
- call packet.destroy() in test

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -2483,6 +2483,80 @@ bare_ffmpeg_packet_set_flags(
   packet->handle->flags = value;
 }
 
+static std::optional<js_arraybuffer_t>
+bare_ffmpeg_packet_get_side_data(
+  js_env_t *env,
+  js_receiver_t,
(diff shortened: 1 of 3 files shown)


kasperisager (changes requested):
`side_data` appears to be an array so we should expose it as such and not attempt to marshall to and from a single blob.

tony-go:
As far as I read, `side_data` is referred as

```
typedef struct AVPacketSideData {
    uint8_t *data;
    size_t   size;
    enum AVPacketSideDataType type;
} AVPacketSideData;
```

I think that we should the struct as it is, maybe this what @kasperisager means by

> so we should expose it as such

tony-go:
Yeah so my recommandation would be to return an `Array` of `SideData` (a new class if needed, as we should be able to access the type easily).

kasperisager:
What do you mean?

Issue #126, opened 2025-09-12
Proper bindings for `CodecContext` and `[Input|Output]FormatContext`


telamon:
I previously worked with the assumption that an `Encoder` can be configured using:
```js
const encoder = new CodecContext(Codec.OPUS.encoder)
const params = new CodecParameters()

params.sampleRate = 44100
params.sampleFormat = sampleFormat.S16

// Apply configuration to Encoder<CodecContext>
params.toContext(encoder) // <-- does not seem to do the expected.

encoder.open()
// encoder is misconfigured
```

But that does not seem to be the case; we have heavy distortion in encoded audio,
most likely the sample format is interpreted wrongly and/or other properties from `params` were not applied.

What i think would be safer to do is to export bindings (getters + setters) for all public members of `AVCodecContext`
That would let me configure the encoder directly without using an intermediary params and be able to _verify that the properties
are truly set_.

telamon:
I figured out the main issue here;

```js
const a = new ffmpeg.CodecParameters()
const context = new CodecContext(...)
a.toContext(context)

context.open() // <-- context has now new values (no bindings)

const b = new ffmpeg.CodecParameters()
b.fromContext(context) // <-- pull post-init props

console.log(a, '!==', b)
```
This is true for both encoders and decoders contexi.

Pulling their props into a codecParameters object after `open()` is a bit of a dirty workaround, not sure if everything is copied.

tony-go:
> But that does not seem to be the case; we have heavy distortion in encoded audio,
most likely the sample format is interpreted wrongly and/or other properties from params were not applied.

Do you have a code snippet to share? @telamon

PR #127, opened 2025-09-12
add properties to CodecContext and [Input|Output]FormatContext

sethvincent:
This is meant to partially address https://github.com/holepunchto/bare-ffmpeg/issues/126 though probably doesn't close it.

The goal with this pr is to expose some additional properties that are useful for managing/understanding CodecContext/InputFormatContext/OutputFormatContext.

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -1149,6 +1149,34 @@ bare_ffmpeg_codec_context_set_extra_data(
 }
 
 static int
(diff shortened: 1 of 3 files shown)


sethvincent:
I'm a little confused about the test failure on darwin-arm64 as it passes on darwin-x64 and ios-arm64. I'll figure that out.

sethvincent:
closing this as i'm about to push a better approach

PR #129, opened 2025-09-14
add avoptions api for codec context

sethvincent:
This pr is meant to help with #126 but probably doesn't close it on its own.


sethvincent:
closing this as i'm about to push a better approach

PR #132, opened 2025-09-16
get supported config from codec

sethvincent:
This provides a way to see what config a codec accepts. It's a newer function in ffmpeg 8.0 and is meant to replace direct access of `codec->supported_framerates` and similar fields.


binding.cc
```diff
@@ -826,6 +826,179 @@ bare_ffmpeg_get_sample_format_name_by_id(js_env_t *env, js_receiver_t, uint32_t
...
+
+  for (size_t i = 0; i < values.size(); i++) {
+    data[i] = values[i];
+  }
+
+  return result;
+}
+
+static js_arraybuffer_t
+bare_ffmpeg_codec_get_supported_frame_rates(
```
tony-go, line 910: My 2 cents regarding that function and the following is that we should only use the one before where we can pass a generic config.

Then we could return an object with a `type` and `values` properties.

{
 type: 'sampleFormat'
 values: ...
}

tony-go (changes requested):

README.md
```diff
@@ -459,6 +459,52 @@ Parameters:
...
+if (pixelFormats) {
+  console.log('Supported pixel formats:', pixelFormats)
+}
+
+const frameRates = codecCtx.getSupportedConfig(
+  ffmpeg.constants.codecConfig.FRAME_RATE
+)
+if (frameRates) {
+  frameRates.forEach((rate) => {
+    console.log(`${rate.numerator}/${rate.denominator} fps`)
```
telamon, line 496: That conversion is built-in to Rational nowadays:
```suggestion
    console.log(`${rate.toNumber()} fps`)
```

binding.cc
telamon: +1 for future pr

telamon (commented):
Nice! these api's reduce amount of magic-numbers for supported codecs.

binding.cc
tony-go: Yeah it could work too. I am not sure on the end.

I would favor a unique function with a `switch`, but it's maybe too personal 😆 

Feel free to skip this comment.

binding.cc
```diff
@@ -826,6 +826,177 @@ bare_ffmpeg_get_sample_format_name_by_id(js_env_t *env, js_receiver_t, uint32_t
...
+  js_arraybuffer_t result;
+  int32_t *data;
+  err = js_create_arraybuffer(env, values.size(), data, result);
+  assert(err == 0);
+
+  for (size_t i = 0; i < values.size(); i++) {
+    data[i] = values[i];
+  }
+
+  return result;
```
tony-go, line 904: Why we don't just return the `values` object? There is a specific reason to use a buffer here?

test/codec-config.js
```diff
@@ -0,0 +1,119 @@
...
+  const codecCtx = new ffmpeg.CodecContext(ffmpeg.Codec.AV1.encoder)
+  const pixFormats = codecCtx.getSupportedConfig(
+    ffmpeg.constants.codecConfig.PIX_FORMAT
+  )
+
+  if (pixFormats) {
+    t.ok(pixFormats instanceof Int32Array, 'returns Int32Array')
+    t.ok(pixFormats.length > 0, 'has pixel formats')
+  } else {
+    t.pass('no specific pixel formats (unbounded)')
```
tony-go, line 14: There is some platforms where it does not work?

tony-go: cc @sethvincent

test/codec-config.js
```diff
@@ -0,0 +1,78 @@
...
+
+test('getSupportedConfig for pixel formats', (t) => {
+  const codecCtx = new ffmpeg.CodecContext(ffmpeg.Codec.AV1.encoder)
+  using pixFormats = codecCtx.getSupportedConfig(
+    ffmpeg.constants.codecConfig.PIX_FORMAT
+  )
+
+  t.ok(pixFormats instanceof Int32Array, 'returns Int32Array')
+  t.ok(pixFormats.length > 0, 'has pixel formats')
+})
```
tony-go, line 12: Sorry I introduce an error in my snippet.

```suggestion
test('getSupportedConfig for pixel formats', (t) => {
  using codecCtx = new ffmpeg.CodecContext(ffmpeg.Codec.AV1.encoder)
  const pixFormats = codecCtx.getSupportedConfig(
    ffmpeg.constants.codecConfig.PIX_FORMAT
  )

  t.ok(pixFormats instanceof Int32Array, 'returns Int32Array')
  t.ok(pixFormats.length > 0, 'has pixel formats')
})
```

We should do the same for other test cases.

binding.cc
```diff
@@ -826,6 +826,164 @@ bare_ffmpeg_get_sample_format_name_by_id(js_env_t *env, js_receiver_t, uint32_t
...
+  js_arraybuffer_span_of_t<bare_ffmpeg_codec_t, 1> codec,
+  int32_t cfg
+) {
+  int err;
+
+  const AVCodecContext *ctx = context->handle;
+  int count = 0;
+  const void *list = nullptr;
+
+  err = avcodec_get_supported_config(ctx, codec->handle, static_cast<AVCodecConfig>(cfg), 0, &list, &count);
```
tony-go, line 843:
```suggestion
  int count = 0;
  const void *list = nullptr;

  err = avcodec_get_supported_config(context->handle, codec->handle, static_cast<AVCodecConfig>(cfg), 0, &list, &count);
```

lib/codec-context.js
```diff
@@ -151,6 +151,58 @@ module.exports = class FFmpegCodecContext {
     return binding.receiveCodecContextFrame(this._handle, frame._handle)
   }
 
+  getSupportedConfig(config) {
+    const constants = require('./constants')
+
+    if (config === constants.codecConfig.FRAME_RATE) {
```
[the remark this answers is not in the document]
sethvincent: Wait there kind of is three cases! lol. But in this situation I feel like the if statements are clear, especially since that last one for SAMPLE_RATE would go in the default block anyway.

binding.cc
```diff
@@ -826,6 +826,163 @@ bare_ffmpeg_get_sample_format_name_by_id(js_env_t *env, js_receiver_t, uint32_t
...
+
+  std::vector<int32_t> values;
+
+  if (count > 0 && list) {
+    const int32_t *int_list = static_cast<const int32_t *>(list);
+    for (int i = 0; i < count; i++) {
+      values.push_back(int_list[i]);
+    }
+    return values;
+  }
```
tony-go, line 858: IIUC the idea between this API is that if nothing is returned then it supports everything?

binding.cc
```diff
@@ -826,6 +826,164 @@ bare_ffmpeg_get_sample_format_name_by_id(js_env_t *env, js_receiver_t, uint32_t
...
+    js_arraybuffer_t result;
+    int32_t *data;
+    err = js_create_arraybuffer(env, 2, data, result);
+    assert(err == 0);
+
+    data[0] = 0;
+    data[1] = 0;
+
+    return result;
+  }
```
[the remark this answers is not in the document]
sethvincent: I was keeping it as close the ffmpeg api as possible to validate that it sent back the sentinel, but maybe that doesn't need to extend into the js side.

I'd be good with switching the return type to an optional. That looks reasonable.

binding.cc
tony-go: I mean `avcodec_get_supported_config`

sethvincent: Yeah, if the list is empty it supports any possible value, which depends on the codec.

It's handled in a separate function but av1 is a good example. There's not set frame rates, instead it's a situational decision.

lib/codec-context.js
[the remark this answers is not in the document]
tony-go: yeah perfect ;)

binding.cc
tony-go: The PR is good otherwise :)

sethvincent: > Is that because we don't have API which exposes all frame rates available?

Right. With av1 for example it supports a wide range that can't be enumerated because it's, like, all the rational numbers (within reason). So instead of having a set of allowed frame rates the constraint is set by whatever other system we're working with and the frame rates it expects. 

And then with channel layouts it's similar but I haven't read as much about it. I think the idea is there are so many possible speaker configurations it would be bad to only allow a certain subset, even though the common ones are 1, 2, 5.1, 7.1, whatever.

tony-go: Okay, feel free to add more details in the doc if needed.

tony-go (approved):

PR #134, opened 2025-09-16
feat(filter-inout): add base class

tony-go:


binding.cc
```diff
@@ -3223,6 +3228,129 @@ bare_ffmpeg_filter_graph_destroy(
...
+  if (filter_inout->handle->next == nullptr) return std::nullopt;
+
+  js_arraybuffer_t handle;
+  bare_ffmpeg_filter_inout_t *next;
+  int err = js_create_arraybuffer(env, next, handle);
+  assert(err == 0);
+
+  next->handle = filter_inout->handle->next;
+
+  return handle;
```
tony-go, line 3341: I am not really sure about that... Should we do a copy. My idea as to use the  same handle. cc @kasperisager

kasperisager: The question is who owns the memory? From the looks of it the memory is owned by the JavaScript wrapper so that should ideally be returned to avoid having the memory disappear from underneath us.

tony-go: > that should ideally be returned

I read "copied". This is what do you mean?

tony-go: I don't have access to the javascript wrapper in that context right? 

But now that I think about it, I think that maybe for `.next` and `.filterContext` I could only keep setters and `.hasNext` and `hasFilterContext` ?

tony-go: Yeah I thought about that but It felt like I was doing a hack (two sources of truth) as I don't access the underlying .data. But now that I re-thing about it, it is wrong as the only accessor of the handle is the class itself, so there is no reason to have a divergence between .next at the C level and the js level.

PR #136, opened 2025-09-16
resizeable `ffmpeg.Samples`

telamon:
There is no guarantee that `decoder.receiveFrame(frame)` produces constant frame-sizes (`nbSamples` may vary).

So I changed the Samples api to contain less state:

- `bare_ffmpeg_samples_init()` removed; replaced with static function `Samples.sizeOf()`
- `new Samples()` constructor takes only the `noAlignment` option.
- `samples.fill(frame)` changes:
  - Reallocates `data` buffer to match size of `frame` when necessary
  - All required values are safely read from `frame`

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -2205,14 +2205,14 @@ bare_ffmpeg_image_get_line_size(
   );
 }
 
-static js_arraybuffer_t
-bare_ffmpeg_samples_init(
+static uint32_t
+bare_ffmpeg_samples_sizeof(
   js_env_t *env,
   js_receiver_t,
   int32_t sample_format,
   int32_t nb_channels,
   int32_t nb_samples,
-  int32_t align
+  bool no_alignment
 ) {
   int err;
 
@@ -2221,7 +2221,7 @@ bare_ffmpeg_samples_init(
     nb_channels,
     nb_samples,
     static_cast<AVSampleFormat>(sample_format),
-    align
+    no_alignment
   );
 
   if (len < 0) {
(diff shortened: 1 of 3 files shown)


test/decode.js
```diff
@@ -153,7 +153,7 @@ function decodeAudio(audio) {
         output.nbSamples = raw.nbSamples
         output.sampleRate = stream.codecParameters.sampleRate
 
-        const samples = new ffmpeg.Samples('S16', 2, output.nbSamples)
+        const samples = new ffmpeg.Samples()
```
tony-go, line 156: I think that it be could interesting for `Sample` to have its own test file where we can test finely each API (like the `noAlignment` case).

binding.cc
```diff
@@ -2205,14 +2205,14 @@ bare_ffmpeg_image_get_line_size(
...
-bare_ffmpeg_samples_init(
+static uint32_t
+bare_ffmpeg_samples_sizeof(
   js_env_t *env,
   js_receiver_t,
   int32_t sample_format,
   int32_t nb_channels,
   int32_t nb_samples,
-  int32_t align
+  bool no_alignment
```
tony-go, line 2215:
```suggestion
  bool has_alignment
```

Wdyt?

telamon: Was a bit confused when i saw it at first, `no_alignment` came from ffmpeg docs https://ffmpeg.org/doxygen/1.0/samplefmt_8c.html#a7368bc4e3a366b688e81938ed55eb06

The variable is vaguely named `align` but it's bounded to 0 = use default alignment, 1 = no alignment.
We can rename + invert it's meaning, but then we have an option that is default `true`, and disabled by passing `false`...

To me that's an anti-pattern, maybe rename to `disableAlignment`, what do you say?

telamon: If we do that then  `align = false` means _enable_ default alignment and `align = true` means _disable_ alignment..  it's an misnomer in ffmpeg.

Honestly i think this option is pretty safe to hardcode to `false` in js and forget / expose it at a later date.
Should i?

test/decode.js
tony-go: I created an issue, I'll see later how can I handle it #141

binding.cc
```diff
@@ -2231,45 +2231,61 @@ bare_ffmpeg_samples_init(
...
-  auto len = av_samples_fill_arrays(
+  if (required < 0) {
+    err = js_throw_error(env, NULL, av_err2str(len));
+    assert(err == 0);
+
+    throw js_pending_exception;
+  }
+
+  if (len < required) {
+    js_throw_errorf(env, NULL, "required at least %zu bytes, got %zu", required, len);
```
kasperisager, line 2266: This is missing an `assert()`.

binding.cc
```diff
@@ -2231,45 +2231,61 @@ bare_ffmpeg_samples_init(
...
   int err;
+  assert(target.size() >= offset + len);
+
+  auto required = av_samples_get_buffer_size(
+    NULL,
+    frame->handle->ch_layout.nb_channels,
+    frame->handle->nb_samples,
+    static_cast<AVSampleFormat>(frame->handle->format),
+    no_alignment
+  );
```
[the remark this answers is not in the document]
kasperisager: I also don't quite see how those could ever diverge in practice, the samples buffer will be reallocated in JavaScript if it's not large enough prior to filling the frame arrays. It seems like unnecessary safety.

kasperisager (changes requested):

lib/samples.js
```diff
@@ -1,49 +1,64 @@
...
+      this._noAlignment
     )
+
+    this._frame = frame
+
+    return res
+  }
+
+  static sizeOf(format, nbChannels, nbSamples, noAlignment = false) {
+    return binding.sizeofSamples(format, nbChannels, nbSamples, noAlignment)
```
kasperisager, line 62:
```suggestion
  static bufferSize(format, nbChannels, nbSamples, noAlignment = false) {
    return binding.samplesBufferSize(format, nbChannels, nbSamples, noAlignment)
```

lib/samples.js
```diff
@@ -1,49 +1,64 @@
...
 
   fill(frame) {
-    binding.fillSamples(
-      this._sampleFormat,
-      this._nbChannels,
-      this._nbSamples,
-      this._align,
+    const { format, channelLayout, nbSamples } = frame
+
+    const len = FFmpegSamples.sizeOf(
```
kasperisager, line 38:
```suggestion
    const len = FFmpegSamples.bufferSize(
```

kasperisager (changes requested):

lib/samples.js
tony-go: yeah I should have reverbate all name changes.

lib/samples.js
```diff
@@ -1,49 +1,64 @@
...
-      this._nbSamples,
-      this._align,
+    const { format, channelLayout, nbSamples } = frame
+
+    const len = FFmpegSamples.bufferSize(
+      format,
+      channelLayout.nbChannels,
+      nbSamples,
+      this._noAlignment
+    )
```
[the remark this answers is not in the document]
kasperisager: Seems fine to expose it as it directly corresponds to `av_samples_get_buffer_size()`.

kasperisager (approved):

lib/samples.js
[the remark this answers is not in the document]
tony-go: The question I raised was not about exposing it or not. But rather do we want to let the user call:
- `bufferSize` and `fill` himself.
- just `fill` which will wrap `bufferSize` (which is the current state)

PR #139, opened 2025-09-18
Rational rescale q

telamon:
- replaced initial JS-impl with accurate `av_rescale_q()` call
- added `Z` export macro for functions don't require  `js_env_t`

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -3369,6 +3369,17 @@ bare_ffmpeg_filter_inout_set_next(
   filter_inout->handle->next = next->handle;
 }
 
+static int64_t
+bare_ffmpeg_rational_rescale_q(
+  int64_t ts,
+  int32_t bq_num,
+  int32_t bq_den,
+  int32_t cq_num,
+  int32_t cq_den
+) {
+  return av_rescale_q(ts, {bq_num, bq_den}, {cq_num, cq_den});
+}
+
 static js_value_t *
 bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   uv_once(&bare_ffmpeg__init_guard, bare_ffmpeg__on_init);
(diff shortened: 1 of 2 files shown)


binding.cc
```diff
@@ -3379,6 +3390,10 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   err = js_set_property<fn>(env, exports, name); \
   assert(err == 0);
 
+#define Z(name, fn) \
+  err = js_set_property<fn,{.scoped = false}>(env, exports, name); \
```
tony-go, line 3394: Why do we set .scoped to false here?

telamon: It's a perf feature in jstl: https://github.com/holepunchto/libjstl/blob/34a7894383a5174353530852416dbb5738a7fafe/include/jstl.h#L2909

It tells jstl to generate a typed_callback without a handle scope, further removing the `js_env_t` arg in callback signature removes all possible overhead from the call.
in short it makes calling `binding.rational_rescale_q()` is almost as cheap as if it was written in js.

Downside is that this feature can only be used for functions that do not interact with the js-state, accept and return primitives (numbers).

Basically limited to math-utils and sizeof calls  in C/C++

binding.cc
```diff
@@ -3379,6 +3390,10 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
   err = js_set_property<fn>(env, exports, name); \
   assert(err == 0);
 
+#define Z(name, fn) \
```
tony-go, line 3393: I would follow the same pattern we follow so far, by overwriting V.

binding.cc
telamon: Affirmative! :fast_forward:

binding.cc
telamon: i did so too first.. but it hurt to break the grouping of exports, i can remove the fastcall entirely if this is a dealbreaker.

tony-go: Let me try to see

kasperisager: Callbacks without `js_env_t` will never implicitly open a handle scope so there's no need to pass `.scoped = false` in the first place.

kasperisager: Put another way, omitting `js_env_t` is the safe way to achieve the fastest of the fast paths.

PR #154, opened 2025-09-26
feat(format-context): expose inputFormat/outputFormat getters

telamon:
When opening a stream the `InputFormat` is auto-detected so added `formatContext.inputFormat` getter.
Similarly in vcr-pipes, we don't access to the original `OutputFormat`, so added `formatContext.outputFormat` getter

@tony-go can you help land this? (we're missing test + docs)


examples/avsynctest-to-webm.js
```diff
@@ -1,16 +1,14 @@
...
   ffmpeg.constants
 
-// ffmpeg.log.level = ffmpeg.log.INFO
-
 const FRAMERATE = 50
 const SAMPLERATE = 48000
 const DURATION = 1
 
-test('write webm', async (t) => {
+async function run() {
```
tony-go, line 11: We should probably reduce the number of asserts @telamon. I just tried to respect the initial work you did for now.

lib/input-format.js
```diff
@@ -19,8 +19,13 @@ switch (Bare.platform) {
...
-  constructor(name = defaultName) {
-    this._handle = binding.initInputFormat(name)
+  constructor(name = defaultName, handle) {
+    if (handle) this._handle = handle
+    else this._handle = binding.initInputFormat(name)
+  }
+
+  static from(handle) {
+    return new FFmpegInputFormat(null, handle)
   }
```
tony-go, line 29: Still not sure we want these as the usage is only internal ...

examples/avsynctest-to-webm.js
telamon: I don't think reducing asserts is a good idea; asserts don't run in production, so asserting as much as possible in dev/native is important.
But that test-file is very old and can be removed / we have better webm-export tests now :+1:

tony-go: > so asserting as much as possible in dev/native is important

We are in JS here right? Or maybe I did not get your point. 

> But that test-file is very old and can be removed / we have better webm-export tests now 👍

`vcr-pipes` tests you mean?

tony-go: I think that for examples we should have one or two assert for the final operation we do. For unit tests I would not be afraid to more tho :) 

> Can be reduced to: https://github.com/holepunchto/vcr-pipes/blob/6ba40ce99fd4746d6dc83f51423413e013042972/test/helpers/index.js#L3-L15


I'll use that then :)

telamon:
Not to derail but thank you for completing the pr, let's merge!

tony-go:
> Not to derail but thank you for completing the pr, let's merge!
https://github.com/holepunchto/bare-ffmpeg/pull/154#discussion_r2388362858

tony-go:
> > Not to derail but thank you for completing the pr, let's merge!
> > [#154 (comment)](https://github.com/holepunchto/bare-ffmpeg/pull/154#discussion_r2388362858)

PR #155, opened 2025-09-27
dumb down time scale ops

telamon:
`packet->time_base` is mostly unused by ffmpeg;
so this patch mirrors the C-api.

### DIFF
--- a/test/packet.js
+++ b/test/packet.js
@@ -127,16 +127,14 @@ test('rescale packet timestamps & timebase', (t) => {
 
   packet.dts = ts
   packet.pts = ts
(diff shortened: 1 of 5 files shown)


test/packet.js
tony-go, line 1: Should we remove the .timeBase setter then ?

telamon: no; we don't know if something uses it; so i'd say keep it for :
a) inspection, "mostly unused" is not same as unused.
b) use it ourselves in higher-levels/abstractions. (think that's what python-ffmpeg did)

PR #156, opened 2025-09-28
enable audio filters

sethvincent:
I went to experiment with audio filters and found I needed to make some changes for that to happen.

This adds the ability to use audio filters and makes a handful of needed changes in various parts of the code.

I'll add comments on the interesting parts.

### DIFF
--- a/binding.cc
+++ b/binding.cc
@@ -2510,6 +2510,33 @@ bare_ffmpeg_samples_fill(
   return res;
 }
 
+static void
+bare_ffmpeg_samples_copy(
+  js_env_t *env,
+  js_receiver_t,
+  js_arraybuffer_span_of_t<bare_ffmpeg_frame_t, 1> dst,
+  js_arraybuffer_span_of_t<bare_ffmpeg_frame_t, 1> src,
+  int32_t dst_offset,
+  int32_t src_offset,
+  int32_t nb_samples
+) {
+  int err = av_samples_copy(
+    dst->handle->data,
+    src->handle->data,
+    dst_offset,
+    src_offset,
+    nb_samples,
+    src->handle->ch_layout.nb_channels,
+    static_cast<AVSampleFormat>(src->handle->format)
+  );
+  if (err < 0) {
+    err = js_throw_error(env, NULL, av_err2str(err));
+    assert(err == 0);
+
+    throw js_pending_exception;
+  }
+}
+
 static int
 bare_ffmpeg_samples_read(
   js_env_t *env,
@@ -2520,7 +2547,6 @@ bare_ffmpeg_samples_read(
   bool no_alignment
 ) {
   int err;
-
   uint8_t *dst_data[8];
   int dst_linesize[4];
 
@@ -3863,6 +3889,7 @@ bare_ffmpeg_exports(js_env_t *env, js_value_t *exports) {
 
   V("samplesBufferSize", bare_ffmpeg_samples_buffer_size)
   V("fillSamples", bare_ffmpeg_samples_fill)
+  V("copySamples", bare_ffmpeg_samples_copy)
   V("readSamples", bare_ffmpeg_samples_read)
 
   V("initPacket", bare_ffmpeg_packet_init)
--- a/lib/samples.js
+++ b/lib/samples.js
@@ -80,6 +80,16 @@ module.exports = class FFmpegSamples {
     )
   }
 
+  copy(frame, dstOffset = 0, srcOffset = 0) {
+    binding.copySamples(
+      frame._handle,
+      this._frame._handle,
+      dstOffset,
+      srcOffset,
+      this.nbSamples
+    )
+  }
+
   static bufferSize(format, nbChannels, nbSamples, noAlignment = false) {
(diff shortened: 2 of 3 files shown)


lib/samples.js
```diff
@@ -58,6 +58,28 @@ module.exports = class FFmpegSamples {
     return res
   }
 
+  read(frame) {
```
sethvincent, line 61: It took me a bit to realize Image had a read method but not Samples. Needed for tests, but probably has actual use cases.

test/samples.js
```diff
@@ -15,7 +15,36 @@ test('Samples sould expose a fill method', (t) => {
...
+  const samplesB = new ffmpeg.Samples()
+  using scratchFrame = makeAudioFrame()
+  samplesB.fill(scratchFrame)
+  samplesB.data.fill(0)
+
+  samplesB.read(sourceFrame)
+
+  t.is(samplesB.data.length, samplesA.data.length, 'same byte length')
+  for (let i = 0; i < view.length; i++) {
+    if (samplesB.data.readInt16LE(i * 2) !== samplesA.data.readInt16LE(i * 2)) {
```
tony-go, line 41: I think that `same-object` used by `brittle` seems to handle `Int16Array` https://github.com/holepunchto/same-object/blob/main/index.js#L58

Maybe we should test if that could work:

```suggestion
  t.alike(samplesB, samplesA)
  // or 
  t.alike(samplesB.data, samplesA.data)
```

lib/samples.js
tony-go: >   Needed for tests, but probably has actual use cases.

Not sure I follow you

binding.cc
```diff
@@ -2491,6 +2510,55 @@ bare_ffmpeg_samples_fill(
...
+
+  err = av_samples_copy(
+    dst_data,
+    frame->handle->data,
+    0,
+    0,
+    frame->handle->nb_samples,
+    frame->handle->ch_layout.nb_channels,
+    static_cast<AVSampleFormat>(frame->handle->format)
+  );
```
kasperisager, line 2551: This should be exposed as a standalone method.

kasperisager (changes requested):

lib/samples.js
sethvincent: Oh, just meant that I only added it to help with writing tests for the filters. I don't know if it has use cases outside of that, but probably.

test/samples.js
sethvincent: aaah! Heck yeah it does. Fixed.

binding.cc
sethvincent: Added `Samples.copy`! I left `bare_ffmpeg_samples_read` as is for now. Should we instead be composing `Samples.fill` and `Samples.copy` in `Samples.read`?

We'll probably want to make equivalent changes in the Image class, yeah? `bare_ffmpeg_samples_read` is nearly the same function as `bare_ffmpeg_image_read`.

kasperisager: Good point, agreed! Smaller methods that can be composed are preferable, with the exception being cases where there's considerable performance to be gained from only round-tripping to native once.

tony-go: What if we also expose `av_samples_fill_arrays` too and then call both on js side, removing `bare_ffmpeg_samples_read`?

Sometimes I wonder if we should keep these elegant wrappers (like `Samples.read`) on userland modules.

lib/samples.js
```diff
@@ -58,6 +58,38 @@ module.exports = class FFmpegSamples {
...
+      channelLayout.nbChannels,
+      nbSamples,
+      this._noAlignment
+    )
+
+    if (!this._data || len !== this._data.length) {
+      this._data = Buffer.allocUnsafe(len)
+    }
+
+    binding.readSamples(
```
tony-go, line 75: I would remove that one and call `fill_arrays` and `copy` here.

binding.cc
sethvincent: It seemed so easy to just use the `fill` and `copy` methods! It was not.

It seemed to work until I got filters involved and found that `bare_ffmpeg_samples_read` was more appropriate for cases where the caller doesn't own the data.

The problem is the difference between frame data the caller owns and frame data being provided by a filter (or similar, I think decoders might be another example?).

When working with the filter-owned data we just want to copy the transformed data out.

In the case of filter-owned data `bare_ffmpeg_samples_read` works because it's not modifying `frame->handle->data`. That's the filter's job!

So, all that to say: I think we should keep `Samples.read` as is for now. It makes sense for the case when a filter is providing the frame's data, and we can use `Samples.fill` in the cases where the caller is providing the frame's data.

We _could_ break this down into additional small functions that make ownership more clear. I'm not sure if the payoff is worth it right now, but reasonable to consider for now or later.

lib/samples.js
sethvincent: see this comment: https://github.com/holepunchto/bare-ffmpeg/pull/156#discussion_r2408000158

binding.cc
kasperisager: That's also a really good point, if the path of least resistance is a separate method for copying non-owned frame data then let's go with that.

kasperisager (approved):

PR #173, opened 2025-10-28
Frame metadata & sidedata, FilterContext options helpers

sethvincent:
This adds metadata and sidedata apis to the Frame class and adds options helpers to FilterContext.


test/filter-context.js
```diff
@@ -5,3 +5,83 @@ test('it should expose a FilterContext class', (t) => {
   const filterCtx = new ffmpeg.FilterContext()
   t.ok(filterCtx)
 })
+
+test('FilterContext exposes runtime option helpers', (t) => {
```
tony-go: I would split the test into two different scenarios. In the first one we assert that get returns something and its type (or that it is the same than default) and in the second we assert the set logic.

In case of regression we could see which part is failing.

test/frame.js
```diff
@@ -114,11 +114,144 @@ test('frame expose a getter for sampleRate', (t) => {
...
+  frame.destroy()
+})
+
+test('Frame.metadata exposes filter-produced entries', (t) => {
+  const { graph, sourceCtx, sinkCtx } = setupAstatsGraph(t)
+
+  const inputFrame = createAudioFrame(0)
+  t.teardown(() => {
+    inputFrame.destroy()
+  })
```
tony-go: `using` could work here.

tony-go:
> Should we fix that in this pull request?

@sethvincent As I said here https://github.com/holepunchto/bare-ffmpeg/pull/173#discussion_r2468745357

> Maybe we should open an issue and discuss how we could generalise with a SideData abstract instead of the dupe PacketSideData/FrameSideData

We could simpliy opening an issue.

tony-go (approved):

PR #174, opened 2025-10-29
fix bare_ffmpeg_filter_graph_parse input/output params

sethvincent:
I tried using multiple inputs and outputs with FilterGraph and found that the params for the binding function needed to be updated to use double pointers to fully support the linked lists.


test/filter-graph.js
```diff
@@ -67,21 +67,79 @@ test('FilterGraph should expose a parse method', (t) => {
...
+
+  using inputs = createFilterInOutChain([
+    [sinkContext0, 'out0'],
+    [sinkContext1, 'out1']
+  ])
+  using outputs = initOutputs(bufferContext)
+
+  t.execution(() => {
+    graph.parse('split=2[out0][out1]', inputs, outputs)
+  })
```
tony-go: According what was added in `lib/filter-graph.js`. It does not look like we have a test that cover that case (having `inputs` or `ouputs` with an `undefined` `_handle`).

tony-go (approved):

PR #179, opened 2025-11-13
feat(hw-device-context): init/destroy (macOS, Linux only)

tony-go:
- I wanted to add the tiniest piece of code that will let me experiment with runtime dependencies on `win32`, `linux` and `android`
- I will add doc later when the API will be cross platform ready

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,11 +10,13 @@ fetch_package("github:holepunchto/libjstl#098664c")
 
 find_port(ffmpeg FEATURES dav1d svt-av1 opus)
 
+set(install_targets)
+
 if(LINUX)
-  add_bare_module(bare_ffmpeg INSTALL TARGET va TARGET va-drm)
-else()
-  add_bare_module(bare_ffmpeg)
+  list(APPEND install_targets TARGET va TARGET va-drm)
 endif()
+
+add_bare_module(bare_ffmpeg INSTALL ${install_targets})
 
 harden(${bare_ffmpeg})
 


CMakeLists.txt
```diff
@@ -8,9 +8,20 @@ project(bare_ffmpeg C CXX)
 
 fetch_package("github:holepunchto/libjstl#098664c")
 
+find_program(
+  pkg-config
+  NAMES pkg-config
+  REQUIRED
+)
+set(CMAKE_PKG_CONFIG "${pkg-config}" CACHE STRING "Path to pkg-config for Meson" FORCE)
```
kasperisager: `cmake-meson` handles this now.

tony-go:
I found that: https://github.com/intel/libva/blob/161f3b51e282cd6d2bc146b2755f14dab73ca014/va/va.c#L383

Maybe just an env variable to configure.

CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ fetch_package("github:holepunchto/libjstl#098664c")
 
 find_port(ffmpeg FEATURES dav1d svt-av1 opus)
 
-add_bare_module(bare_ffmpeg)
+if(LINUX)
+  add_bare_module(bare_ffmpeg INSTALL TARGET va TARGET va-drm)
```
kasperisager: Split out `TARGET va TARGET va-drm` into a separate list and then do `add_bare_module(bare_ffmpeg INSTALL ${the_list})` so `add_bare_module(bare_ffmpeg)` is only present once in the code.

test/hw-device-context.js
```diff
@@ -0,0 +1,44 @@
+const os = require('bare-os')
+const test = require('brittle')
+const ffmpeg = require('..')
+const process = require('bare-process')
+
+if (os.platform() === 'darwin') {
+  test('HWDeviceContext should instantiate with device type', (t) => {
```
kasperisager:
```suggestion
  test('HWDeviceContext should instantiate with device type', { skip: os.platform() !== 'darwin' }, (t) => {
```

The same applies to the other cases; mark them skipped instead of conditionally declaring them.

kasperisager (changes requested):

PR #180, opened 2025-11-18
feat(hw-device-context): Init API fro linux with VAAPI

tony-go:


tony-go:
Still WIP, I just try to trigger the CI.

PR #212, opened 2026-03-25
refactor(format-context): add transfer method

tony-go:
## Context

I saw users shooting themselves into their feet because of this used case:

```js
  using io = new ffmpeg.IOContext(image)
  using inputFormatContext = new ffmpeg.InputFormatContext(io)
```

... which provokes a double free in the current implementation.

## Proposal

Add a `.transfer` method to transfer ownership

### DIFF
--- a/lib/format-context.js
+++ b/lib/format-context.js
@@ -8,9 +8,7 @@ const Dictionary = require('./dictionary')
 
 class FFmpegFormatContext {
   constructor(io) {
-    if (io) {
-      this._io = io.transfer()
-    }
+    this._io = io ? io.transfer() : null
     this._streams = []
   }
 
--- a/lib/format-context.js
+++ b/lib/format-context.js
@@ -8,7 +8,11 @@ const Dictionary = require('./dictionary')
 
 class FFmpegFormatContext {
   constructor(io) {
-    this._io = io
+    if (io) {
+      const ownedIO = new IOContext(null, null)
+      io.transfer(ownedIO)
+      this._io = ownedIO
+    }
     this._streams = []
   }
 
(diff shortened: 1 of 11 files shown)


lib/io-context.js
```diff
@@ -27,7 +32,14 @@ module.exports = class FFmpegIOContext {
...
 
   destroy() {
-    binding.destroyIOContext(this._handle)
+    if (this._handle) {
+      binding.destroyIOContext(this._handle)
+      this._handle = null
+    }
+  }
+
+  transfer(to) {
```
kasperisager:
```suggestion
  transfer() {
    const to = new IOContext(null, null)
```

lib/io-context.js
```diff
@@ -27,7 +32,14 @@ module.exports = class FFmpegIOContext {
...
+    to._handle = this._handle
     this._handle = null
```
kasperisager, line 44:
```suggestion
    this._handle = null
    return to
```

lib/format-context.js
```diff
@@ -8,7 +8,11 @@ const Dictionary = require('./dictionary')
 
 class FFmpegFormatContext {
   constructor(io) {
-    this._io = io
+    if (io) {
+      const ownedIO = new IOContext(null, null)
+      io.transfer(ownedIO)
+      this._io = ownedIO
```
kasperisager:
```suggestion
      this._io = io.transfer()
```

kasperisager (changes requested):

lib/format-context.js
```diff
@@ -8,7 +8,9 @@ const Dictionary = require('./dictionary')
 
 class FFmpegFormatContext {
   constructor(io) {
-    this._io = io
+    if (io) {
+      this._io = io.transfer()
+    }
```
kasperisager:
```suggestion
    this._io = io ? io.transfer() : null
```

That way the object shape is static.

kasperisager (changes requested):

kasperisager (approved):

PR #220, opened 2026-07-13
fix(format-context): leaks

tony-go:


lib/format-context.js
```diff
@@ -91,15 +91,26 @@ exports.InputFormatContext = class FFmpegInputFormatContext extends FFmpegFormat
...
-      this._handle = binding.openInputFormatContextWithIO(this._io._handle)
+      try {
+        this._handle = binding.openInputFormatContextWithIO(this._io._handle)
+      } catch (err) {
+        // The native open failed (e.g. a JS IO callback threw). The io handle
+        // was already transferred into this._io, so free it here before
+        // rethrowing, otherwise the AVIOContext and its callback references
+        // leak (the caller's original io is now a no-op empty handle).
+        super.destroy()
+        throw err
```
kasperisager:
```suggestion
      } finally {
        super.destroy()
```

kasperisager (changes requested):

kasperisager (dismissed):

lib/format-context.js
tony-go: The comment leads to a fake conclusion the destroy has to be done:
- in the destroy (end of this ressource lifecycle)
- in a catch when the binding function thrown

tony-go: cc @kasperisager

kasperisager (approved):

PR #232, opened 2026-09-16
feat: enable vulkan for win32

SudhansuBandha:
Closes #188


SudhansuBandha:
Hi @kasperisager ,
Do let me know whether the proposed changes meet expectations or not. I am happy to make further changes.

cmake/ports/ffmpeg/port.cmake
```diff
@@ -119,14 +119,24 @@ elseif(WIN32)
     --enable-d3d12va
     --enable-dxva2
     --enable-mediafoundation
+    --enable-vulkan
+    --host-extralibs=
```
kasperisager, line 123: What's the purpose of an empty `--host-extralibs`?

cmake/ports/ffmpeg/port.cmake
```diff
@@ -119,14 +119,24 @@ elseif(WIN32)
...
   )
+
```
kasperisager, line 125:
```suggestion
```

test/hw-device-context.js
```diff
@@ -47,3 +47,76 @@ test('HWDeviceContext.from should return null for null handle', (t) => {
...
+      t.ok(frame.height > 0, 'decoded frame has height')
+
+      break
+    }
+
+    if (decoded) break
+  }
+
+  t.ok(decoded, 'decoded at least one Vulkan frame')
+})
```
kasperisager, line 122:
```suggestion
})

```

kasperisager (changes requested):
Make sure to run Prettier as well, there are formatting inconsistencies that will show up as lint error once we approve the workflow run.

cmake/ports/ffmpeg/port.cmake
SudhansuBandha: Build for windows was failing as `FFmpeg's` configure step was adding `HOSTEXTRALIBS=-lm` which was looking for `m.lib` on Windows. This was resulting in linking failure. Hence I added this additional configuration to address this linking failure. 

Is this line not required?

kasperisager: I don't see where `-lm` would come from on Windows, that's the `libc` math library. Try removing it and then we'll see if it reproduces on the Actions runner.

SudhansuBandha: Updated PR as suggested!!

SudhansuBandha: I believe this issue `HOSTEXTRALIBS=-lm` is resolved post upgrade of `FFmpeg` version in `main`

SudhansuBandha:
I have installed Vulkan SDK on my Windows for this. Do we need to perform the same in the CI environment while trying to build?

kasperisager:
We do, yes. One thing we need to confirm is that this doesn't add a strong dependency on the Vulkan SDK for the user of the module.

SudhansuBandha:
Let me validate this and address any issues if found. Then hopefully CI failure should get resolved

